PR #1, opened 2019-12-19
Allow to provide Optional param value in Uri.param

abdolence:
Hey, thanks for a great library.

It would be great if we could provide optional values to the `Uri.param` like

```
request.
    param("test", Some(testValue))
```


adamw:
Thanks! released as 1.0.0-RC2

PR #3, opened 2020-01-24
Make scheme required in uri interpolation

### LINKED ISSUE #2: `Uri.parse("")` and `uri""` return a successful Uri instance ("http://")
It seems unintuitive to accept an empty string and default to an invalid Uri.
Here's a quick example:

```scala
import sttp.model.Uri

// This returns
// Either[String, Uri] = Right(
//   Uri("http", None, Segment("", sttp.model.Uri$HostEncoding$$$Lambda$2312/576222471@75ae4a1f), None, Vector(), Vector(), None)
// )
val uri1 = Uri.parse("")
uri1.map(_.toJavaUri) // java.net.URISyntaxException: Expected authority at index 7: http://

val uri2 = uri"".toJavaUri // Same exception as above
```

If this behavior is unintentional, I'd be happy to fix this in a PR.

kusamakura:
Closes #2.

This is a breaking change for those who rely on the uri interpolation to provide `http` as the default scheme.

Work required for softwaremill/sttp#396 seems to be a lot more involved, so I plan to tackle that issue in another PR.

### DIFF
--- a/core/shared/src/main/scala/sttp/model/UriInterpolator.scala
+++ b/core/shared/src/main/scala/sttp/model/UriInterpolator.scala
@@ -338,7 +338,8 @@ object UriInterpolator {
     case object Scheme extends UriBuilder {
       override def fromTokens(u: Uri, t: Vector[Token]): (Uri, Vector[Token]) = {
         split(t, Set[Token](SchemeEnd)) match {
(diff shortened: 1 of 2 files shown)


adamw:
Released in 1.0.0-RC3, thanks!

PR #4, opened 2020-01-29
Add support for unknown attributes in a cookie

bilal-fazlani:
More information can be found here: https://github.com/softwaremill/sttp/issues/415

With the changes in this PR, now when sttp client encounters a cookie with an unknown attribute, it does not fail. It simply adds it to a collection of "otherAttributes"


core/shared/src/main/scala/sttp/model/Cookie.scala
```diff
@@ -71,7 +71,8 @@ case class CookieValueWithMeta private (
     domain: Option[String],
     path: Option[String],
     secure: Boolean,
-    httpOnly: Boolean
+    httpOnly: Boolean,
+    otherAttributes: Map[String, Option[String]]
```
adamw: It seems that the official terminology is "directives" (see e.g. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie), so maybe `otherDirectives`?

core/shared/src/main/scala/sttp/model/Cookie.scala
```diff
@@ -262,4 +282,10 @@ object CookieWithMeta {
           case Failure(_) => Left(())
         }
     }
+
+  private implicit class StringInterpolations(sc: StringContext) {
+    def ci = new {
```
adamw: I think we should make this a proper type. Returning a structural type causes this to be invoked using reflection at run-time, which is an unnecessary performance penatly.

But a nice pattern!

core/shared/src/test/scala/sttp/model/CookieTests.scala
```diff
@@ -0,0 +1,44 @@
+package sttp.model
```
adamw: these tests should go to the jvm sources, otherwise js tests fail because of some problems with `Instant` etc. which are unavailable in JS. I have to fix it one day, but that's another issue :)

adamw:
Released in 1.0.0-RC6

PR #8, opened 2020-03-08
Handling mailformed (in respect to date) weekday names

FunctorPrototype:
I've faced with next _expires value_ ```Thu, 01-Jan-2114 00:00:10 GMT``` It seems to look good but 01-Jan-2114 is Monday not Thursday.  Using DateTimeFormatter's parse method causes exception due wrong weekday name. I'm not sure if [standard](https://tools.ietf.org/html/rfc2616#section-3.3.1) contains explicit requrement to week day name relation to date, but above _expires value_ string is correct in sense of _expires value_ grammar. ```java.net.HttpCookie``` for that string simply doesnt parse _expires value_. There is no any exceptions in ```java.net.HttpCookie.parse```. In turn Google Chrome doesnt drop such cookie or _expires value_ .That looks logically becase weekday name is some kind of redundant information. I think Google Chrome behaviour can be considered as canonical in the case. So I've tried to adapt Rfc850 datetime parsing to that Chrome behaviour in this PR


adamw:
Released in 1.0.2, thanks!

PR #10, opened 2020-04-02
Provide default values for `CookieValueWithMeta`

jan0sch:
Add default values for the `safeApply` and `unsafeApply` methods on
`CookieValueWithMeta` to ease usage.

Also `.gitignore` contains now entries to ignore files generated by
Bloop and Metals.


adamw:
Released in 1.0.3, thanks!

PR #25, opened 2020-09-21
Add convenience .accept method to Header companion

kalanzai:
When adding the "Accept" header to a request it is convenient
to be able to use the defined MediaTypes rather than having
to construct the string yourself


adamw:
Released in 1.2.0-RC4, thanks!

PR #45, opened 2020-11-27
Fixed form-data params encoding

pinguinson:
Found couple of issues while fixing https://github.com/softwaremill/sttp/issues/773:
1) form-data disposition parameters' values should be enquoted
2) some HTTP servers expect `name` to be the first parameter

### DIFF
--- a/core/src/main/scala/sttp/model/Part.scala
+++ b/core/src/main/scala/sttp/model/Part.scala
@@ -41,10 +41,12 @@ case class Part[+T](
 
   def contentDispositionHeaderValue: String = {
     def encode(s: String): String = new String(s.getBytes("utf-8"), "iso-8859-1")
-    "form-data; " + dispositionParams.map { case (k, v) => s"$k=${encode(v)}" }.mkString("; ")
+    "form-data; " + dispositionParams.map { case (k, v) => s"""$k="${encode(v)}"""" }.mkString("; ")
   }
 
-  def dispositionParams: Map[String, String] = otherDispositionParams + (NameDispositionParam -> name)
+  // some servers require 'name' disposition parameter to be first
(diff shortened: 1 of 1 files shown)


core/src/main/scala/sttp/model/Part.scala
```diff
@@ -41,10 +41,12 @@ case class Part[+T](
...
   def contentDispositionHeaderValue: String = {
     def encode(s: String): String = new String(s.getBytes("utf-8"), "iso-8859-1")
-    "form-data; " + dispositionParams.map { case (k, v) => s"$k=${encode(v)}" }.mkString("; ")
+    "form-data; " + dispositionParams.map { case (k, v) => s"""$k="${encode(v)}"""" }.mkString("; ")
   }
 
-  def dispositionParams: Map[String, String] = otherDispositionParams + (NameDispositionParam -> name)
+  // some servers require 'name' disposition parameter to be first
+  // see: https://stackoverflow.com/questions/20261088/certain-order-of-fields-in-content-disposition-with-jersey-client
+  def dispositionParams: Map[String, String] = Map(NameDispositionParam -> name) ++ otherDispositionParams
```
adamw, line 49: If we want to guarantee order, then we'll need to use a different type. A `Map` can arbitarily reorder elements regardless of the order in which they are added.

Maybe this should return a `LinkedMap`, or simply a `List[(String, String)]` (a list of tuples)?

pinguinson: Good call! Added `orderedDispositionParams` method to avoid changing existing method's signature and thus breaking backwards compatibility.

adamw:
Release in 1.2.0-RC6, thank you!

adamw:
Forgot to write, I changed the method names so that they follow the conventions in `Uri`, hope you don't mind :)

PR #51, opened 2020-12-23
Support relative URIs

adamw:
The scheme and authority are now optional in an URI.


adamw:
Released in https://github.com/softwaremill/sttp-model/releases/tag/v1.2.0-RC10

PR #53, opened 2020-12-28
Support encoding of 3+ UTF-8 byte characters

satabin:
For such characters, the java `Char` which contains only the first 2
bytes cannot be converted to a valid string which result in a string
containing a single byte `0x3F`.

By iterating over the UTF-8 bytes instead of 2-byte characters, we can
handle all different cases, treating the allowed characters as the
single byte characters they are.

This can occur with some APIs, such as the Discord API, which requires the reaction emoji to be part of the path in URL-encoded form.


adamw:
Thank you! Released in https://github.com/softwaremill/sttp-model/releases/tag/v1.2.0-RC10

Issue #59, opened 2021-01-26
Relative Uri path has an empty item as head


kgston:
Thanks for the quick fix in 1.2.1! 
I bumped into this (maybe cosmetic?) issue where the `path` list of path segments has a blank entry as the first object when the uri is in the form `/a/b`. If you tried to combine the relative path with another absolute path, you would end up with a resulting Uri with `//` where it was concatenated. It is fine for the case where the Uri is parsed from `a/b`, but if the relative Uri only has a single path segment like `a`, it would not parse properly without the leading `/` denoting it is a relative path.

```scala
import sttp.model._

val test = Uri.unsafeParse("/a")
val endpoint = Uri.unsafeParse("http://test.com/abc")
println(test.path) // List(, a)
println(endpoint.addPath(test.path)) // http://test.com/abc//a
```

adamw:
Hm good catch and good question!

The problem here is that we need to both represent path-relative (e.g. `a/b/c`) and path-absolute (e.g. `/a/b/c`) relative URIs. WIth absolute URIs this is simple, as there are no relative paths. Here we've got this additional complexity.

An open question is how to solve this :). We've need to somehow let the user know (at least through documentation) what to expect from `.path` and how to recognize path-absolute and path-relative URIs.

One way is introducing a `pathAbsolute: Boolean` flag to the `URI` class. But then ... wouldn't it be surprising that both `Uri.unsafeParse("/a").path` and `Uri.unsafeParse("a").path` return the same? The paths are different in both cases.

I'm open to suggestions and ideas :)

adamw:
One more note: probably the example is just a demo of the problem, but if you've got a base URI & want to make it more specific, I think either replacing the path with `.withPath` or `.resolve` could work better?

Adding an absolute path to an URI with a path already defined - not sure what the result here should be :)

kgston:
Hmm, I didn't know about the difference in the path-relative & path-absolute implementation. I could work with just the path-relative format, except that I also noticed that paths with a single segment like `Uri.unsafeParse("path1").path` would be `List()` and `path1` would end up being parsed as the `scheme`. Not sure if this is a bug or intended, but if single segment relative path can be done, it would fit my use case.

One use case that I was working on was with AWS API Gateway. They have this format where it ends with a path segment as part of the base URI. For example `https://123456789.execute-api.ap-northeast-1.amazonaws.com/main`. So I would have 2 instances of that for different environments, e.g.
[...]

PR #61, opened 2021-01-28
Path ADT, .addPath improvements

adamw:


adamw:
@kgston Related to our discussion in #59, see what you think about the design. I think it solves the problems described there.

@jadlr This also solves the `.addPath` problem you mentioned on gitter

jadlr:
Great! Thank you so much for the quick response!

kgston:
Thanks for looking into this, I think it looks good!

PR #70, opened 2021-02-24
ability to add charset to multipart part content type header

dkarwacki:


core/src/main/scala/sttp/model/Part.scala
```diff
@@ -49,6 +49,13 @@ case class Part[+T](
   // some servers require 'name' disposition parameter to be first
   // see: https://stackoverflow.com/questions/20261088/certain-order-of-fields-in-content-disposition-with-jersey-client
   def dispositionParamsSeq: Seq[(String, String)] = (NameDispositionParam -> name) :: otherDispositionParams.toList
+
+  def addCharsetToContentType(charset: String): Part[T] = {
+    contentType.map(ct => if (!ct.contains("charset")) s"$ct; charset=$charset" else ct) match {
```
adamw, line 54: this check should first lower-case `ct`

adamw: I think it would be simpler to do:

```
contentType match {
  case Some(ct) if !ct.toLowerCase.contains("charset") => ...
   case _ =>  this
}
```

dkarwacki: corrected

PR #76, opened 2021-03-17
Accept header parsing MediaType extensions

kubinio123:


core/src/main/scala/sttp/model/MediaType.scala
```diff
@@ -13,6 +13,20 @@ case class MediaType(mainType: String, subType: String, charset: Option[String]
   def charset(c: String): MediaType = copy(charset = Some(c))
   def noCharset: MediaType = copy(charset = None)
 
+  def matches(other: MediaType): Boolean =
+    (this, other) match {
+      case (MediaType("*", _, _), MediaType("*", _, _))                                                                           => true
```
adamw, line 18: the `*` could be a (private) constant

core/src/main/scala/sttp/model/MediaType.scala
```diff
@@ -13,6 +13,20 @@ case class MediaType(mainType: String, subType: String, charset: Option[String]
   def charset(c: String): MediaType = copy(charset = Some(c))
   def noCharset: MediaType = copy(charset = None)
 
+  def matches(other: MediaType): Boolean =
```
adamw, line 16: Maybe we could move the parameter parsing to this class. Following https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types, the media type can have arbitrary parameters. Currently we support only one `charset` - but here we need to add support also for `q` parameters.

So let's add a `otherParameter: Map[String, String]` to the case class fields, and fill it during media type parsing? (+ proper serialisation)

core/src/main/scala/sttp/model/headers/Accept.scala
```diff
@@ -0,0 +1,41 @@
+package sttp.model.headers
+
+import sttp.model.{Header, HeaderNames, MediaType}
+
+import scala.collection.immutable.Seq
+
+private[model] sealed trait Accept[A] {
+  def parse(headers: Seq[Header]): Seq[(A, Float)]
```
adamw, line 8: I think we could have better representation for accept-like headers than a sequence of tuples - it's quite hard to guess what this is, and even harder to expected that these will be sorted

Maybe we could add a `case class Accept[T](sortedValues: Seq[(T, Float)])`. `Accept` (for mediatype), `accept-charset` would then be specialisations of this type. Possibly we could also have here a function to pick the best matching option?

kubinio123: Regarding this and comment above:

Maybe we could add `q` as a separate field to `MediaType` and avoid sequence with floats handling it like so:
* add a method to `HasHeaders` trait called `accepts` returning `Seq[MediaType]`
* this method will
** parse `Accept` header to `Seq[MediaType]` using `MediaType.parse()`
** parse `Accept-Charset` to `Seq[String, Float]`
** merge it together
for example:
`Accept: text/csv;q=1, text/plain;q=0.9`
`Accept-Charset: utf;q=1, iso;q=0.9`
will result in four `MediaType's`:
`MediaType(text, csv, utf, 1)`,
`MediaType(text, csv, iso, 0.9)`,
`MediaType(text, plain, utf, 0.9)`, // takes smaller q
`MediaType(text, plain, iso, 0.9)`

and sort it by `q` and position of media type/charset in headers to handle same `q`.

Then whole content negotiation will be simplified to taking available `MediaType's` from codecs and collecting first match from `headers.accept()`.

What do you think?

adamw: Merging & sorting sounds good, though I would be wary of adding `q` as a parameter to `MediaType` - it is only used for `Accepts`, while media type itself is used in other contexts as well (most commonly in `Content-Type`). That's why I'd like to keep it as simple as possible, without singling out any additional parameters (above `charset`). So I'd opt to keep all other parameters is a map (empty by default)

core/src/main/scala/sttp/model/MediaType.scala
```diff
@@ -1,19 +1,55 @@
...
+) {
   def charset(c: Charset): MediaType = charset(c.name())
   def charset(c: String): MediaType = copy(charset = Some(c))
   def noCharset: MediaType = copy(charset = None)
 
-  override def toString: String = s"$mainType/$subType" + charset.fold("")(c => s"; charset=$c")
+  def matches(other: MediaType): Boolean = {
+    val typeMatches = (this, other) match {
+      case (MediaType(Wildcard, Wildcard, _, _), _)                                                              => true
+      case (_, MediaType(Wildcard, Wildcard, _, _))                                                              => true
```
adamw, line 23: so both `text/plain` matches `*/*` and `*/*` matches `text/plain`?

I'm reading "x matches y" as "x matches the pattern y" so only the first should be true. If you understand "matches" differently, maybe we need another verb?

adamw: http4s uses "satisfies", and it is also a one-way relation
akka-http uses "matches" but I think the other way round than I described above

core/src/main/scala/sttp/model/MediaType.scala
```diff
@@ -1,19 +1,55 @@
...
+      case _ => false
+    }
+
+    if (typeMatches) {
+      if (this.isCharsetAny || other.isCharsetAny) true
+      else this.charset.map(_.toLowerCase) == other.charset.map(_.toLowerCase)
+    } else false
+  }
+
+  def matchesExact(other: MediaType): Boolean = {
```
adamw, line 40: maybe we can have a `def isRange` - which would check if this media type represents a range, of it's "exact".

Then instead of having a `matchesExact` clients could use `isRange` & `matches`

core/src/main/scala/sttp/model/MediaType.scala
```diff
@@ -1,19 +1,55 @@
...
+    this.mainType.equalsIgnoreCase(other.mainType) &&
+    this.subType.equalsIgnoreCase(other.subType) &&
+    this.charset.map(_.toLowerCase) == other.charset.map(_.toLowerCase)
+  }
+
+  def isMainTypeAny: Boolean = mainType == Wildcard
+  def isSubTypeAny: Boolean = subType == Wildcard
+  def isTypeAny: Boolean = isMainTypeAny && isSubTypeAny
+  def isCharsetAny: Boolean = charset.forall(_ == Wildcard)
```
adamw, line 49: can charset be *?

core/src/main/scala/sttp/model/MediaType.scala
```diff
@@ -1,19 +1,55 @@
...
+    } else false
+  }
+
+  def matchesExact(other: MediaType): Boolean = {
+    this.mainType.equalsIgnoreCase(other.mainType) &&
+    this.subType.equalsIgnoreCase(other.subType) &&
+    this.charset.map(_.toLowerCase) == other.charset.map(_.toLowerCase)
+  }
+
+  def isMainTypeAny: Boolean = mainType == Wildcard
```
adamw, line 46: do we need/want to expose those methods to the public?

core/src/main/scala/sttp/model/headers/Accepts.scala
```diff
@@ -0,0 +1,55 @@
+package sttp.model.headers
+
+import sttp.model.internal.Patterns
+import sttp.model.internal.Validate._
+import sttp.model.{Header, HeaderNames, MediaType}
+
+import scala.collection.immutable.{Map, Seq}
+
+private[model] object Accepts {
```
adamw, line 9: I think I'd lean towards making this public, but removing `HasHeaders.accepts`. This would make it similar to how we handle e.g. `CacheDirectives`. An `accepts` method in both requests & responses (they all inherit from `HasHeaders`) might be confusing, and this is only used for a narrow set of server-side use-cases

core/src/main/scala/sttp/model/headers/Accepts.scala
```diff
@@ -0,0 +1,55 @@
...
+  def unsafeParse(headers: Seq[Header]): Seq[MediaType] = {
```
adamw, line 10: we should also have a "safe" `parse` variant

core/src/main/scala/sttp/model/internal/Patterns.scala
```diff
@@ -0,0 +1,57 @@
+package sttp.model.internal
+
+import java.util.regex.Pattern
+import scala.util.matching.Regex
+
+private[model] object Patterns {
+
+  private val Token = "([a-zA-Z0-9-!#$%&'*+.^_`{|}~]+)"
```
adamw, line 8: do these somehow reference the RFC? might be good to add a link here, where this comes from

core/src/main/scala/sttp/model/internal/Patterns.scala
```diff
@@ -0,0 +1,57 @@
...
+  private val Quoted = "\"([^\"]*)\""
+
+  val Type: Pattern = Pattern.compile(s"$Token")
+  val TypeSubtype: Pattern = Pattern.compile(s"$Token/$Token")
+  val Parameter: Pattern = Pattern.compile(s";\\s*(?:$Token=(?:$Token|$Quoted))?")
+
+  val QValue: Regex = "[0-1](.\\d\\d?\\d?)?".r
+  val WhiteSpaces: String = "\\s+"
+
+  def parseParameters(t: String, offset: Int): Either[String, Map[String, String]] = {
```
adamw, line 18: are "parameters" a concept in the http spec that applies to a wider area, or are these only media type parameters? If the latter, let's call this `parseMediaTypeParameters`, as that's what this is

core/src/main/scala/sttp/model/MediaType.scala
kubinio123: In context of Accept-Charset header yes - https://tools.ietf.org/html/rfc7231#section-5.3.3

adamw: but a `MediaType(*, *, charset = Some(*))` which serializes to `*/*; charset=*` is incorrect, right? If so, we cannot use such incorrect representations

core/src/main/scala/sttp/model/MediaType.scala
kubinio123: Yes, both match. Another verb that comes to my mind is `fits`. To avoid this two-way relation in instance method maybe we could move that to MediaType object method which takes two params?

adamw: I think `matches` is good, but I'd use a one-way relation. It's just a matter of picking the direction ;)

An instance method would also need to specify which argument is the pattern and which the "matched" media type

core/src/main/scala/sttp/model/MediaType.scala
kubinio123: In context of usage in tapir this method is not needed, only `isTypeAny` and `isCharsetAny` is used but can be replaced with pattern matching . Probably it won't be used by public since picking media type will be handled internally. So I will remove it making `Wildcard` constant public

core/src/main/scala/sttp/model/MediaType.scala
kubinio123: So I will go for other being the pattern here

core/src/main/scala/sttp/model/internal/Patterns.scala
kubinio123: renamed

kubinio123: Params are not mentioned in other header docs

adamw: ok, thx, so let's keep the names specific then :)

core/src/main/scala/sttp/model/headers/Accepts.scala
kubinio123: added

core/src/main/scala/sttp/model/internal/Patterns.scala
kubinio123: yes, I found a mention about that. Can contain any char except some reserved "separators"

core/src/main/scala/sttp/model/MediaType.scala
```diff
@@ -1,19 +1,39 @@
...
+) {
   def charset(c: Charset): MediaType = charset(c.name())
   def charset(c: String): MediaType = copy(charset = Some(c))
   def noCharset: MediaType = copy(charset = None)
 
-  override def toString: String = s"$mainType/$subType" + charset.fold("")(c => s"; charset=$c")
+  def matches(range: ContentTypeRange): Boolean = {
+    def charsetMatches: Boolean =
+      if (range.charset == Wildcard) true
+      else this.charset.map(_.toLowerCase).contains(range.charset.toLowerCase)
```
adamw, line 24: not `==`?

core/src/main/scala/sttp/model/ContentTypeRange.scala
```diff
@@ -0,0 +1,8 @@
+package sttp.model
+
+case class ContentTypeRange(mainType: String, subType: String, charset: String)
+
+object ContentTypeRange {
+  val Wildcard = "*"
+  val AnyContentTypeRange: ContentTypeRange = ContentTypeRange(Wildcard, Wildcard, Wildcard)
```
adamw, line 7: maybe simply `AnyRange` (`Any` is taken by the type ;) )?

core/src/main/scala/sttp/model/MediaType.scala
```diff
@@ -1,19 +1,39 @@
...
-
-import sttp.model.internal.Validate
+import sttp.model.ContentTypeRange.Wildcard
 import sttp.model.internal.Rfc2616._
+import sttp.model.internal.Validate._
+import sttp.model.internal.{Patterns, Validate}
 
-case class MediaType(mainType: String, subType: String, charset: Option[String] = None) {
+import java.nio.charset.Charset
+import scala.collection.immutable.{Map, Seq}
```
adamw, line 9: [very minor] you shouldn't need to import immutable map? only seq has a different default impl

core/src/main/scala/sttp/model/MediaType.scala
```diff
@@ -22,68 +42,61 @@ object MediaType extends MediaTypes {
...
+    }
   }
 
   def unsafeParse(s: String): MediaType = parse(s).getOrThrow
+
+  def bestMatch(mediaTypes: Seq[MediaType], ranges: Seq[ContentTypeRange]): Option[MediaType] = {
+    mediaTypes
+      .map(mt => mt -> ranges.indexWhere(mt.matches))
+      .filter({ case (_, i) => i != NotFoundIndex }) // not acceptable
+      .sortBy({ case (_, i) => i })
```
adamw, line 94: maybe `minBy` would work better here?

core/src/main/scala/sttp/model/MediaType.scala
kubinio123: could be, but followed by `getOrElse`

core/src/main/scala/sttp/model/headers/Accepts.scala
```diff
@@ -0,0 +1,94 @@
...
+object Accepts {
+  def parse(headers: Seq[Header]): Either[String, Seq[ContentTypeRange]] =
+    (parseAcceptHeader(headers), parseAcceptCharsetHeader(headers)) match {
+      case (Right(mts), Right(chs))         => Right(toContentTypeRanges(mts, chs))
+      case (Left(errorMts), Left(errorChs)) => Left(s"$errorMts\n$errorChs")
+      case (Left(error), _)                 => Left(error)
+      case (_, Left(error))                 => Left(error)
+    }
+
+  def unsafeParse(headers: Seq[Header]): Seq[ContentTypeRange] =
```
adamw, line 19: can't we implement this using `parse(headers).getOrThrow`? maybe we could avoid duplicating some code

core/src/main/scala/sttp/model/MediaType.scala
adamw: ah sorry, that's a contains on an option, not on a string, that's totally fine then

core/src/main/scala/sttp/model/MediaType.scala
kubinio123: right it's not needed

core/src/main/scala/sttp/model/headers/Accepts.scala
```diff
@@ -0,0 +1,94 @@
...
+import sttp.model.ContentTypeRange._
+import sttp.model.internal.Patterns
+import sttp.model.internal.Validate._
+import sttp.model.{ContentTypeRange, Header, HeaderNames, MediaType}
+
+import scala.collection.immutable.{Map, Seq}
+
+object Accepts {
+  def parse(headers: Seq[Header]): Either[String, Seq[ContentTypeRange]] =
+    (parseAcceptHeader(headers), parseAcceptCharsetHeader(headers)) match {
```
adamw, line 12: I think the following:
```
for {
 mts <- parseAcceptHeader
 chs <- parseCharset
} yield toContentType(...)
```

should do the same, without the need to unpack the errors

kubinio123: yes, did that at first but  Either in not a monad in earlier scala versions :( yields errors

adamw: ah right ... well, what can we do :)

core/src/main/scala/sttp/model/headers/Accepts.scala
```diff
@@ -0,0 +1,94 @@
...
+    (mediaTypes, charsets) match {
+      case (Nil, Nil) => Seq(AnyContentTypeRange)
+      case (Nil, chs) =>
+        chs.sortBy({ case (_, q) => -q }).map { case (ch, _) => ContentTypeRange(Wildcard, Wildcard, ch) }
+      case (mts, Nil) =>
+        mts.sortBy({ case (_, q) => -q }).map { case (mt, _) => ContentTypeRange(mt.mainType, mt.subType, Wildcard) }
+      case (mts, chs) =>
+        val merged = mts.flatMap { case (mt, mtQ) =>
+          // if Accept-Charset is defined then any other charset specified in Accept header in not acceptable
+          chs.map { case (ch, chQ) => (mt, ch) -> math.min(mtQ, chQ) }
```
adamw, line 38: can we do a `min` here? is there a defined precedence of media type & charset q-values?

core/src/main/scala/sttp/model/internal/Patterns.scala
```diff
@@ -0,0 +1,60 @@
...
+
+private[model] object Patterns {
+
+  /** token can consist any char except separators defined in rfc
+    * https://tools.ietf.org/html/rfc2616#section-2.2
+    */
+  private val Token = "([a-zA-Z0-9-!#$%&'*+.^_`{|}~]+)"
+  private val Quoted = "\"([^\"]*)\""
+
+  val Type: Pattern = Pattern.compile(s"$Token")
```
adamw, line 14: that's only for media types? Maybe also use `MediaType` here?

core/src/main/scala/sttp/model/MediaType.scala
kubinio123: actually I could but I would have to pattern match first to check if not empty. There's minByOption which does that but it's `value minByOption is not a member of scala.collection.immutable.Seq`

core/src/test/scalajvm/sttp/model/headers/AcceptsTest.scala
```diff
@@ -0,0 +1,121 @@
...
+    Accepts.parse(Seq(invalidAccept)) shouldBe Left(
+      """Parameter is not formatted correctly: "=q=1" for: "text/html;=q=1""""
+    )
+
+    val invalidAcceptCharset = Header(HeaderNames.AcceptCharset, "utf-8;=a=1")
+    Accepts.parse(Seq(invalidAcceptCharset)) shouldBe Left(
+      """Parameter is not formatted correctly: "=a=1" for: "utf-8;=a=1""""
+    )
+
+    Accepts.parse(Seq(invalidAccept, invalidAcceptCharset)) shouldBe Left(
```
adamw, line 116: we should probably also have a test when `q` is not a float

core/src/main/scala/sttp/model/MediaType.scala
adamw: ah ... well, maybe let's do the pattern match, sorting here isn't really needed as we need the smallest value only

core/src/main/scala/sttp/model/headers/Accepts.scala
kubinio123: I didn't found a defined precedence. But it seems correct to handle that this way.

If client accepts type A with qA=1 and charset C with qC=0.5 means that he accept A with charset C but since charset C has lower quality resulting content type will also be with lower quality - :d masło maślane but it's hard to describe this as example.

Akka does that too.

adamw: ok, thx :)

core/src/main/scala/sttp/model/internal/Patterns.scala
kubinio123: `Type` is also used for matching charsets from `Accept-Charset`. Name `TokenPattern` would be appropriate but no meaningful name comes to mind for `TypeSubtype` then.

core/src/test/scalajvm/sttp/model/headers/AcceptsTest.scala
kubinio123: added

core/src/main/scala/sttp/model/headers/Accepts.scala
kubinio123: right, I missed that

PR #82, opened 2021-04-16
Add Sec-WebSocket-Protocol header definition

barthorre:


adamw:
Thanks, should be released soon in 1.4.2: https://github.com/softwaremill/sttp-model/actions/runs/755487623

PR #84, opened 2021-04-20
q value regex fix

kubinio123:
So, the `.2` is fixed for `q` value, but `text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2` from issue is still invalid, since single `*` is not valid in Accept header, subtype is missing

### DIFF
--- a/core/src/main/scala/sttp/model/MediaType.scala
+++ b/core/src/main/scala/sttp/model/MediaType.scala
@@ -6,7 +6,6 @@ import sttp.model.internal.Validate._
 import sttp.model.internal.{Patterns, Validate}
 
 import java.nio.charset.Charset
-import scala.collection.immutable.Seq
 
 case class MediaType(
     mainType: String,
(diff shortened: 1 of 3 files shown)


adamw:
Well, maybe it's invalid, but we have to live with it ;) This is a header that is sent by default by the http client built into jdk, so we have to support it. See: https://bugs.openjdk.java.net/browse/JDK-8163921

kubinio123:
so it's about that, thought that `.2` is the issue here :) so one more adjustment

adamw:
I think both were issues :)

core/src/main/scala/sttp/model/MediaType.scala
```diff
@@ -72,8 +71,11 @@ object MediaType extends MediaTypes {
     if (!typeSubtype.lookingAt()) {
       return Left(s"""No subtype found for: "$t"""")
     }
-    val mainType = typeSubtype.group(1).toLowerCase
-    val subType = typeSubtype.group(2).toLowerCase
+
+    val (mainType, subType) = (typeSubtype.group(1), typeSubtype.group(2), typeSubtype.group(3)) match {
+      case (null, null, Wildcard) => (Wildcard, Wildcard)
```
adamw, line 76: maybe add a comment what these `null`s mean and why do we have this special case?

kubinio123: added

PR #118, opened 2021-09-24
Implemented model for Content-Range header

bartekzylinski:


core/src/main/scala/sttp/model/headers/RangeHeader.scala
```diff
@@ -0,0 +1,53 @@
+package sttp.model.headers
+
+import sttp.model.internal.Validate.RichEither
+import sttp.model.{ContentRangeUnits, HeaderNames}
+
+import scala.util.Try
+
+case class RangeHeader(start: Option[Int], end: Option[Int], unit: String) {
```
adamw, line 8: maybe just `Range` since we have just `ContentRange` (without `...Header`)?

core/src/main/scala/sttp/model/headers/ContentRange.scala
```diff
@@ -0,0 +1,51 @@
...
+      if (contentRange.range.isDefined && contentRange.size.isDefined) {
+        val isRangeAndSizeValid = contentRange
+          .range.exists(range => contentRange.size.exists(size => range._1 < size & range._2 <= size))
+        isRangeValid && !isSyntaxInvalid && isRangeAndSizeValid
+      } else isRangeValid && !isSyntaxInvalid
+    } else !isSyntaxInvalid
+  }
+
+  def unsafeApply(s: String): ContentRange = safeApply(s).getOrThrow
+  def safeApply(s: String): Either[String, ContentRange] = parse(s)
```
adamw, line 50: `*apply` creates the model from already parsed values. Here we don't do any validation on the data, so the auto-generated case class `apply` is enough.

core/src/main/scala/sttp/model/headers/ContentRange.scala
```diff
@@ -0,0 +1,51 @@
...
+      val isRangeValid = contentRange.range.exists(r => r._1 < r._2)
+      if (contentRange.range.isDefined && contentRange.size.isDefined) {
+        val isRangeAndSizeValid = contentRange
+          .range.exists(range => contentRange.size.exists(size => range._1 < size & range._2 <= size))
+        isRangeValid && !isSyntaxInvalid && isRangeAndSizeValid
+      } else isRangeValid && !isSyntaxInvalid
+    } else !isSyntaxInvalid
+  }
+
+  def unsafeApply(s: String): ContentRange = safeApply(s).getOrThrow
```
adamw, line 49: we should have an `unsafeParse`, though :)

core/src/main/scala/sttp/model/headers/ContentRange.scala
```diff
@@ -0,0 +1,51 @@
...
+object ContentRange {
+
+  def parse(str: String): Either[String, ContentRange] = {
+    Try(parseString(str))
+      .filter(isValid)
+      .toEither.left
+      .map(_.getMessage)
+  }
+
+  private def parseString(str: String): ContentRange = {
```
adamw, line 21: it would be better to implement `parseUnsafe` in terms of `parse`, not the other way round - avoiding exceptions if possible (that is, if the user uses `parse`)

core/src/main/scala/sttp/model/headers/RangeHeader.scala
adamw: is `RangeHeader(None, Some(10), "bytes")` valid?

adamw: and it should be `Long`, no?

bartekzylinski: 1 ok
2 yes you can do that it should be interpreted as 10 last bytes of file
3 it should be I will do it after fixing compilation errors

core/src/main/scala/sttp/model/headers/ContentRange.scala
bartekzylinski: I probably should change this method name it was aim to be simple util method not connected to class "API"

core/src/main/scala/sttp/model/headers/ContentRange.scala
```diff
@@ -0,0 +1,57 @@
...
+}
+
+object ContentRange {
+
+  def parse(str: String): Either[String, ContentRange] =
+    str.trim.split(" ") match {
+      case Array(unit, s) =>
+        s.split("/") match {
+          case Array(possibleRange, possibleSize) => processString(unit, possibleRange, possibleSize)
+          case _                                  => Left("Unable to parse incorrect string: %s".format(s))
```
adamw, line 18: let's give more context in the error message - e.g. `expected range in the format: x/y, but got: s`

core/src/main/scala/sttp/model/headers/ContentRange.scala
```diff
@@ -0,0 +1,57 @@
...
+    val contentRange = ContentRange(unit, range, size)
+    if (isValid(contentRange)) Right(contentRange)
+    else Left("Invalid Content-Range")
+  }
+
+  private def isValid(contentRange: ContentRange): Boolean = {
+    val isSyntaxInvalid = contentRange.range.isEmpty && contentRange.size.isEmpty
+    if (contentRange.range.isDefined) {
+      val isRangeValid = contentRange.range.exists(r => r._1 < r._2)
+      if (contentRange.range.isDefined && contentRange.size.isDefined) {
```
adamw, line 39: instead of `.isDefined`, it's usually better to pattern-match, as you get access to the values and you can't really mess up the branches ;) Here you could do

```(contentRange.range, contentRange.size) match {
  case (Some(...), Some(...)) =>
  case _ =>
}

File: core/src/main/scala/sttp/model/headers/ContentRange.scala [hunk 7]
```diff
@@ -0,0 +1,57 @@
...
+          case Array(possibleRange, possibleSize) => processString(unit, possibleRange, possibleSize)
+          case _                                  => Left("Unable to parse incorrect string: %s".format(s))
+        }
+      case _ => Left("Unable to parse incorrect string: %s".format(str))
+    }
+
+  private def processString(unit: String, possibleRange: String, possibleSize: String): Either[String, ContentRange] = {
+    val range = possibleRange.split("-") match {
+      case Array("*")        => None
+      case Array(start, end) => Some((start.toLong, end.toLong))
```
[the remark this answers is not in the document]
adamw: would be good to use `.toLongOption`, with different impls for Scala 2.12- (using .toInt and Try) and 2.13+ (using the built-in function)

core/src/main/scala-2.13/sttp/model/internal/ParseUtils.scala
```diff
@@ -0,0 +1,10 @@
+package sttp.model.internal
+
+import scala.util.Try
+
+object ParseUtils {
+
+  def toLongOption(s: String): Option[Long] =
+    Try(s.toLong).toOption
```
adamw, line 8: the whole point of splitting this is to use `.toLongOption` here :)

core/src/main/scala/sttp/model/headers/ContentRange.scala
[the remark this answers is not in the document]
adamw: and here we should use our `ParseUtils` :)

core/src/main/scala/sttp/model/headers/Range.scala
```diff
@@ -0,0 +1,73 @@
...
+
+import sttp.model.internal.ParseUtils
+import sttp.model.internal.Validate.RichEither
+import sttp.model.{ContentRangeUnits, HeaderNames}
+
+import scala.annotation.tailrec
+
+case class Range(start: Option[Long], end: Option[Long], unit: String) {
+
+  val range: RangeValue = RangeValue(start, end)
```
adamw, line 11: I'm not convinced of the utility of the `RangeValue` class. Can't we just put its functionality here? `Range.range.toContentRange` looks weird ;-)

PR #132, opened 2021-10-27
Uri.addQuerySegments

ddworak:
Just for consistency with `addParams` and `addPathSegments`, which enable batch updates to the URI. I can also add `Seq` variant if you'd like one, but this is enough for my own convenience ;)


adamw:
Thanks :)

PR #138, opened 2021-11-15
Added support for Accept-Encoding header

Pask423:


core/src/main/scala/sttp/model/headers/AcceptEncoding.scala
```diff
@@ -0,0 +1,53 @@
+package sttp.model.headers
+
+import sttp.model.internal.Validate.RichEither
+
+case class AcceptEncoding(compressionAlgorithm: String, weight: Option[BigDecimal]) {
```
adamw, line 5: According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding I think the right name for `compressionAlgorithm` would be `contentEncoding`

Also, there might be multiple content encodings specified, each with an optional weight (and I think that's often the case). So maybe sth like `case class AcceptEncoding(contentEncodings: List[Weighted[String]])`?

core/src/main/scala/sttp/model/headers/AcceptEncoding.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  def unsafeParse(s: String): List[AcceptEncoding] = parse(s).map(_.getOrThrow)
+
+  def unsafeApply(compressionAlgorithm: String, weight: Option[BigDecimal]): AcceptEncoding =
+    safeApply(compressionAlgorithm, weight).getOrThrow
+
+  def safeApply(compressionAlgorithm: String, weight: Option[BigDecimal]): Either[String, AcceptEncoding] = {
+    val encoding = AcceptEncoding(compressionAlgorithm, weight)
+    if (isValid(encoding)) Right(encoding)
+    else Left("Invalid Content Range")
```
adamw, line 51: invalid encoding? or do we in fact validate the weight?

README.md
```diff
@@ -36,6 +36,7 @@ Available model classes include:
    * `ServerSentEvent`  
 * header values:
    * `Accepts`
+   * `Accept-Encoding`
```
adamw, line 39: We should also add a factory method to `Header` which accepts the `AcceptEncoding` value

core/src/main/scala/sttp/model/headers/AcceptEncoding.scala
```diff
@@ -0,0 +1,66 @@
+package sttp.model.headers
+
+import sttp.model.internal.Validate.RichEither
+
+import scala.annotation.tailrec
+
+case class Encoding(encoding: String, weight: Option[BigDecimal]) {
```
adamw, line 7: I think I'd move it inside the `AcceptEncoding` object - or will it be used by also header values as well? If not, then the weighted-header seems like a detail of the accept encoding header, not a "global" concept

core/src/main/scala/sttp/model/headers/AcceptEncoding.scala
```diff
@@ -0,0 +1,66 @@
...
+  override def toString: String = encodings.map(_.toString).mkString(",")
+}
+
+object AcceptEncoding {
+
+  def parse(str: String): Either[String, AcceptEncoding] = {
+    val encodings = processString(str, List.empty)
+    if (encodings.forall(isValid) && encodings.nonEmpty) Right(AcceptEncoding(encodings.reverse))
+    else if (encodings.isEmpty)
+      Left("Expected Accept-Encoding in the format: \"deflate or gzip;q=1.0\", but got: %s".format(str))
```
adamw, line 22: you can't have an `or` in the header value, right? ;)

core/src/main/scala/sttp/model/headers/AcceptEncoding.scala
```diff
@@ -0,0 +1,66 @@
...
+        weight.split("=") match {
+          case Array(_, value) => Encoding(algorithm, Some(BigDecimal(value)))
+          case _               => Encoding("", None)
+        }
+      case Array(algorithm) => Encoding(algorithm, None)
+      case _                => Encoding("", None)
+    }
+  }
+
+  private def isValid(acceptEncoding: Encoding): Boolean = {
```
adamw, line 49: maybe we can return a more descriptive error message saying what's wrong?

core/src/main/scala/sttp/model/headers/AcceptEncoding.scala
Pask423: yeah should fix this

PR #142, opened 2021-12-05
Implementation of Authenticate Challenge parse

Pask423:


adamw:
Can you merge in master so that we can verify that the change is binary-compatible?

core/src/test/scala/sttp/model/headers/WWWAuthenticateChallengeTest.scala
```diff
@@ -3,25 +3,43 @@ package sttp.model.headers
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
 
+import scala.collection.immutable.ListMap
+
 class WWWAuthenticateChallengeTest extends AnyFlatSpec with Matchers {
 
   it should "properly serialise a basic header value" in {
-    WWWAuthenticateChallenge.basic.toString shouldBe "Basic"
+    WWWAuthenticateChallenge.parseSingle("Basic") shouldBe Right(WWWAuthenticateChallenge("Basic"))
```
adamw, line 11: I think the serialisation tests should remain intact, but we should add new parsing tests?

core/src/main/scala/sttp/model/AuthenticationSchemes.scala
```diff
@@ -0,0 +1,80 @@
+package sttp.model
+
+import scala.collection.immutable.ListMap
+
+object AuthenticationSchemes {
+
+  val RealmParam = "realm"
```
adamw, line 7: these params are not auth schemes :)

core/src/main/scala/sttp/model/AuthenticationSchemes.scala
```diff
@@ -0,0 +1,80 @@
...
+  val CharsetParam = "charset"
+  val BasicScheme: String = Basic.name
+  val BearerScheme: String = Bearer.name
+  val DigestScheme: String = Digest.name
+  val supportedSchems: List[String] = List(Basic.name, Bearer.name, Digest.name)
+
+}
+
+object Basic {
```
adamw, line 16: we definitely need to move those to some objects to avoid cluttering the sttp.model namespace with `Basic` etc

core/src/main/scala/sttp/model/AuthenticationSchemes.scala
```diff
@@ -0,0 +1,77 @@
+package sttp.model
+
+import scala.collection.immutable.ListMap
+
+object AuthenticationSchemes {
```
adamw, line 5: is this & `AuthenticationChallengeParams` be going to be used elsewhere as well? That is, for other model classes

Pask423: Hard to say but probably not

adamw: It should be possible to determine if that's the case - is there any other place in the HTTP spec - another header, more specifically - that uses these concepts? Browsing MDN might be enough to check that

core/src/main/scala/sttp/model/headers/WWWAuthenticateChallenge.scala
```diff
@@ -22,14 +24,57 @@ case class WWWAuthenticateChallenge(scheme: String, params: ListMap[String, Stri
...
+
+  val RealmParam: String = "realm"
+  val CharsetParam: String = "charset"
+  val BasicScheme: String = Basic.name
+  val BearerScheme: String = Bearer.name
+
+  def parseSingle(str: String): Either[String, WWWAuthenticateChallenge] = {
+    str.trim.replaceFirst(" ", "_").split("_") match {
+      case Array(x, possibleParams) =>
+        if (AuthenticationSchemes.supportedSchemes.forall(possibleParams.contains))
```
adamw, line 36: why forall? I think 2 are enought to signal an error

core/src/main/scala/sttp/model/headers/WWWAuthenticateChallenge.scala
```diff
@@ -22,14 +24,57 @@ case class WWWAuthenticateChallenge(scheme: String, params: ListMap[String, Stri
...
+          Left(s"Multiple challenges in single header not supported but found in: $str")
+        else {
+          val params = creteParamsMap(possibleParams.trim)
+          x.trim match {
+            case BasicScheme =>
+              if (params.size > AuthenticationSchemes.Basic.maxParametersCount) Left(s"To much params for Basic in: $possibleParams")
+              else Right(WWWAuthenticateChallenge(AuthenticationSchemes.Basic.name, AuthenticationSchemes.Basic.getParams(params)))
+            case BearerScheme =>
+              if (params.size > AuthenticationSchemes.Bearer.maxParametersCount) Left(s"To much params for Bearer in: $possibleParams")
+              else Right(WWWAuthenticateChallenge(AuthenticationSchemes.Bearer.name, AuthenticationSchemes.Bearer.getParams(params)))
```
adamw, line 46: we are doing very partial checking here - verifying that there's not too many parameters, but not checking if there are no extra unsupported ones (is that an error according to the spec?), and if the required ones are present (does the spec specify required params?)

core/src/main/scala/sttp/model/AuthenticationSchemes.scala
adamw: Ah ok, that's useful then. Still I would simplify and expose as little as possible. E.g. the `getParams` methods seems like something that should be private to the parser. On the other hand, centralising the names of the schemes in this object sounds reasonable.

core/src/main/scala/sttp/model/headers/WWWAuthenticateChallenge.scala
Pask423: According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate (for Basic and Digest) and https://www.rfc-editor.org/rfc/rfc6750.html#page-7 (for Bearer) most of params is optional. I will have to fix Digest, which have some required params, handling but others should work fine. All schemas have well described list of params so i thought that if our string contains more then that we should stop parsing such header.

core/src/main/scala/sttp/model/AuthenticationSchemes.scala
```diff
@@ -0,0 +1,85 @@
...
+
+  val supportedSchemes: List[String] = List(Basic.name, Bearer.name, Digest.name)
```
adamw, line 7: can we make this private or model-private? Everything we expose needs to remain for the whole life of 1.x

core/src/main/scala/sttp/model/AuthenticationSchemes.scala
```diff
@@ -0,0 +1,85 @@
...
+
+  object Basic {
+
+    val maxParametersCount = 2
```
adamw, line 11: same here - I'd only expose what's really necessary for the users

core/src/main/scala/sttp/model/AuthenticationSchemes.scala
```diff
@@ -0,0 +1,85 @@
...
+    val name = "Basic"
+
+    private val realm: String = "realm"
+    private val charset: String = "charset"
+
+    def getParams(params: Map[String, String]): ListMap[String, String] =
```
adamw, line 17: and here - that's internal for validation

core/src/main/scala/sttp/model/AuthenticationSchemes.scala
```diff
@@ -0,0 +1,85 @@
...
+    private val opaque: String = "opaque"
+    private val stale: String = "stale"
+    private val algorithm: String = "algorithm"
+    private val qop: String = "qop"
+    private val qopValues = List("auth", "auth-int")
+    private val charset: String = "charset"
+    private val userhash: String = "userhash"
+
+    def paramsValid(params: Map[String, String]): Boolean = {
+      val matchingParamsNumber = params.size <= maxParametersCount
```
adamw, line 63: either we are strict & checking that there are only allowed parameters, or we allow extra parameters and validate them. Checking that the number of params doesn't exceed some value doesn't translate to anything that's in the spec

core/src/main/scala/sttp/model/AuthenticationSchemes.scala
```diff
@@ -0,0 +1,85 @@
...
+      val containsNonce = params.contains(nonce)
+      val containsOpaque = params.contains(opaque)
+      val qopValue = params.getOrElse(qop, "")
+      val qopValueMatch = qopValues.exists(_.equals(qopValue))
+      matchingParamsNumber && containsNonce && containsOpaque && qopValueMatch
```
adamw, line 68: so nonce, opaque and qop are required parameters?

core/src/main/scala/sttp/model/headers/WWWAuthenticateChallenge.scala
```diff
@@ -22,14 +24,79 @@ case class WWWAuthenticateChallenge(scheme: String, params: ListMap[String, Stri
...
 
 object WWWAuthenticateChallenge {
-  val RealmParam = "realm"
-  val CharsetParam = "charset"
-  val BasicScheme = "Basic"
-  val BearerScheme = "Bearer"
+
+  val RealmParam: String = "realm"
+  val CharsetParam: String = "charset"
+  val BasicScheme: String = Basic.name
```
adamw, line 30: maybe these should be `@deprecated` as we have these constants defined elsewhere?

core/src/main/scala/sttp/model/AuthenticationSchemes.scala
Pask423: yep according to spec for Digest

adamw: ok maybe we can be more user-friendly and simply validate this with proper error messages, such as "missing nonce parameter"

core/src/main/scala/sttp/model/headers/WWWAuthenticateChallenge.scala
adamw: 1. the `@deprecated` has a message field which allows adding user-friendly deprecation notices, such as what to use instead
2. is the intention of `AuthenticationSchemes.Basic` to be public or internal API? Currently its visibility is scoped to sttp-model only, so we can't point users to use it instead of this field

Pask423: Hmm probably more internal one I think that we want to expose only "interface" from `WWWAuthenticateChallenge` without making creation of particular challenge possible in any other way

PR #175, opened 2022-02-18
Change Header.unapply to return Some

### LINKED ISSUE #174: Header.unapply return type makes pattern matches non-exhaustive
`Header.unapply` [returns an `Option`](https://github.com/softwaremill/sttp-model/blob/e76986ca7687935f62dc22bfbafc5c05726a1a52/core/src/main/scala/sttp/model/Header.scala#L61). From the compiler's point of view, this means that it can potentially fail - meaning a pattern match like this
```scala
header match {
  case Header(_, _) => ???
}
```
is non-exhaustive, and triggers the corresponding compiler lint. 

We can fix this by changing the return type of `unapply` to `Some[(String, String)]` - see [this Scastie](https://scastie.scala-lang.org/qunCaMBzQLi2yPf3ankiLg).

nigredo-tori:
Fixes #174

### DIFF
--- a/core/src/main/scala/sttp/model/Header.scala
+++ b/core/src/main/scala/sttp/model/Header.scala
@@ -58,7 +58,7 @@ class Header(val name: String, val value: String) {
 /** For a description of the behavior of `apply`, `safeApply` and `unsafeApply` methods, see [[sttp.model]].
   */
 object Header {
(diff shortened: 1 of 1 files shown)


adamw:
Thanks, but it seems that unfortunately this is a binary-incompatible change. Will have to wait for a major release.

PR #189, opened 2022-05-10
Add tags to RequestMetadata

sergio-margale:


adamw:
On further thought, `tags` should not constitute request metadata: they're an sttp-client specific thing, and other usages (e.g. Tapir) might not be tag-aware.

PR #194, opened 2022-05-23
Empty string handling

Pask423:
closes https://github.com/softwaremill/sttp/issues/1072 and first case from https://github.com/softwaremill/sttp/issues/396


adamw:
Thanks! :)

PR #195, opened 2022-05-23
Corner case handling

Pask423:
closes second case from https://github.com/softwaremill/sttp/issues/396

@adamw should we also handle `http:/` as incorrect uri somehow ?


core/src/test/scala/sttp/model/UriInterpolatorTests.scala
```diff
@@ -217,6 +217,7 @@ class UriInterpolatorTests extends AnyFunSuite with Matchers {
     ("uri with incorrect hostname", () => uri"https://./search", "incorrect hostname"),
     ("empty string as uri", () => uri" ", "empty string is not valid uri"),
     ("empty uri", () => uri"", "empty string is not valid uri"),
+    ("corner case", () => uri"https:/ /uri", "empty string is not valid host and port")
```
adamw, line 220: this looks like a valid relative uri so I wouldn't reject it, even if it's most probably a mistake; but if we start adding such corner-cases we'll end up with 1000 if-s ;)

Issue #196, opened 2022-06-07
Improve performance of UriInterpolator


danilbykov:
I often see stack traces of UriInterpolator inside JFR performance profiles. It looks like parsing and interpolation of URLs takes considerable amount of CPU.

I did dirty changes (not ready for pull request yet) inside UriInterpolator and got 3x or 4x performance improvement.

Here is result of my benchmarks
```
[info] Benchmark                                   Mode  Cnt       Score   Error  Units
[info] InterpolatorBenchmark.longUrlCurrent       thrpt    2   82954,476          ops/s
[info] InterpolatorBenchmark.longUrlFast          thrpt    2  236963,125          ops/s
[info] InterpolatorBenchmark.simpleUrlCurrent     thrpt    2  111629,426          ops/s
[info] InterpolatorBenchmark.simpleUrlFast        thrpt    2  413596,768          ops/s
[info] InterpolatorBenchmark.urlWithQueryCurrent  thrpt    2   53065,576          ops/s
[info] InterpolatorBenchmark.urlWithQueryFast     thrpt    2  191607,313          ops/s
```

Basically I did two main changes:
1. Inside Tokenizers I collect tokens in java.util.ArrayList instead of Vector
1. For UriBuilder I wrote structure similar to View from scala2.13 and create Uri without Vectors

So I mainly removed allocations of Vectors. Of course code looks more complex but I think performance gain is considerable. What do you think? Are you interested in these changes?

adamw:
Interesting! :) Yes, if you could create a PR, please do

PR #197, opened 2022-06-13
Improve performance of UriInterpolator

danilbykov:
This is pull request for https://github.com/softwaremill/sttp-model/issues/196.

I split my pull request into three commits:

1. First commit replaces `Vector` in `UriBuilder`s by `sttp.model.internal.ArrayView`. This allows to build `Uri` using one array (without allocations many Vectors inside each `UriBuilder`)
2. Second commit replaces `Vector` in `Tokenizer`s by `java.util.ArrayList`. Again this allows to tokenize string using one or two `ArrayList`.
3. Third commit replaces `Set` by `sttp.model.internal.FastCharSet`. Call `Set.contains(char)` leads to boxing of character which is slower than call `FastCharSet.contains(char)`. In this commit I also replaced regular expressions by custom code based on `FastCharSet`. This also improved performance.

Here is results of JMH benchmarks:
```
Without any changes
[info] Benchmark                           Mode  Cnt  Score   Error  Units
[info] InterpolatorBenchmark.longUrl       avgt    2  0,195           s/op
[info] InterpolatorBenchmark.simpleUrl     avgt    2  0,108           s/op
[info] InterpolatorBenchmark.urlWithQuery  avgt    2  0,305           s/op

With ArrayView
[info] Benchmark                           Mode  Cnt  Score   Error  Units
[info] InterpolatorBenchmark.longUrl       avgt    2  0,146           s/op
[info] InterpolatorBenchmark.simpleUrl     avgt    2  0,079           s/op
[info] InterpolatorBenchmark.urlWithQuery  avgt    2  0,241           s/op

With ArrayList
[info] Benchmark                           Mode  Cnt  Score   Error  Units
[info] InterpolatorBenchmark.longUrl       avgt    2  0,094           s/op
[info] InterpolatorBenchmark.simpleUrl     avgt    2  0,059           s/op
[info] InterpolatorBenchmark.urlWithQuery  avgt    2  0,165           s/op

With FastCharSet
[info] Benchmark                           Mode  Cnt  Score   Error  Units
[info] InterpolatorBenchmark.longUrl       avgt    2  0,052           s/op
[info] InterpolatorBenchmark.simpleUrl     avgt    2  0,030           s/op
[info] InterpolatorBenchmark.urlWithQuery  avgt    2  0,079           s/op
```

### DIFF
--- a/benchmark/src/main/scala/sttp/model/InterpolatorBenchmark.scala
+++ b/benchmark/src/main/scala/sttp/model/InterpolatorBenchmark.scala
@@ -1,0 +1,37 @@
+package sttp.model
+
+import java.util.concurrent.TimeUnit
+
+import org.openjdk.jmh.annotations._
+import sttp.model.Uri._
+
+@BenchmarkMode(Array(Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@Warmup(iterations = 4, time = 3, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 2, time = 20, timeUnit = TimeUnit.SECONDS)
+@Fork(value = 1, jvmArgsAppend = Array("-Xms1G", "-Xmx1G", "-XX:+UnlockDiagnosticVMOptions", "-XX:+DebugNonSafepoints"))
+@State(Scope.Benchmark)
+class InterpolatorBenchmark {
+
+  @Benchmark
+  def simpleUrl: Int = {
+    uri"http://example.com/some/path".path.length
+  }
+
+  @Benchmark
+  def longUrl: Int = {
+    val var1 = "path"
+    val var2 = "another"
+    val var3 = "other"
+    val var4 = "zxcv"
+    uri"http://example.com/some/$var1/$var2/$var3/$var4/123".path.length
+  }
+
+  @Benchmark
+  def urlWithQuery: Int = {
+    val value3 = "value3"
+    val value4 = "value4"
+    val value5 = "value5"
+    uri"http://example.com/some?key1=value1&key2=value2&key3=$value3&key4=$value4&key5=$value5".path.length
+  }
+}
--- a/build.sbt
+++ b/build.sbt
@@ -97,3 +97,15 @@ lazy val core = (projectMatrix in file("core"))
     scalaVersions = scala2 ++ scala3,
     settings = commonNativeSettings
   )
+
+lazy val benchmark = (projectMatrix in file("benchmark"))
+  .settings(
+    name := "benchmark",
+    publish / skip := true
+  )
+  .jvmPlatform(
+    scalaVersions = scala2,
+    settings = commonJvmSettings
+  )
+  .dependsOn(core)
+  .enablePlugins(JmhPlugin)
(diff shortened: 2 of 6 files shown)


Pask423:
For me it look more or less ok. Unfortunately it breaks mima so we will have to wait probably till sttp v4

danilbykov:
I fixed most of incompatibilities by adding intermediate traits `SourceCompatibleTokenizer` and `SourceCompatibleUriBuilder`.

But there is still one incompatible change due to parameter with default argument in this method
```scala
    private def tokenizeTerminatedFragment(
        s: String,
        current: Tokenizer,
        terminators: Set[Char],
        separatorsToTokens: Map[Char, Token],
        separatorsEscape: Option[(Char, Char)] = None
    ): (Tokenizer, Vector[Token]) = {
```
which adds method `default$5` in class file
```scala
public class sttp.model.UriInterpolator$Tokenizer$ {
  ...
  public scala.Option<scala.Tuple2<java.lang.Object, java.lang.Object>> sttp$model$UriInterpolator$Tokenizer$$tokenizeTerminatedFragment$default$5();
  ...
}
```
Parameters with default arguments [should never](https://github.com/jatcwang/binary-compatibility-guide#dont-adding-parameters-with-default-values-to-methods) used in library code.

But it doesn't look as serious incompatibility. I don't see how removing this method may lead to LinkageError unless someone directly access this method.

adamw:
Maybe we'll first release a version of sttp-model with the tokenizer etc. marked as package-private, which should keep binary compatibility, but break source compatibility, and we'll see if somebody complains? If not, it should be safe to make them private

danilbykov:
I like idea making `Tokenizer`, `UriBuilder` and `Token`s private. I made these changes.

> Finally, a question - is there a particular reason to use java.util.List instead of Scala's ListBuffer, for example?

`ListBuffer` is based on `List` so index-based operations are slow. But I found out that scala library has also `scala.collection.mutable.ArrayBuffer` which is similar to `ArrayList` from java. Do you like `ArrayBuffer` more than `ArrayList`? I can replace probably. I don't see methods of `ArrayBuffer` which is necessary in current code but maybe they will be needed in future.

> Could I also ask for one more favor - some comments to the chosen design. There's a mutable list, a var, etc. maybe you could write some short info on what these mutable represent during the process of uri parsing.

Do you mean comments in code or comments in pull request?

There is one mutable list (ArrayList now) which is used to collect tokens. There are couple functions where fast index-based operations are needed. For example, Authority.tokenize replaces token for IPv6 address or `removeEmptyTokensAroundExp`. So it limits list of appropriate collections for this accumulator.

Once all tokens are collected in `ArrayList`, I copy these tokens in plain Array and wrap this Array in `ArrayView`. Copying is needed here because initial list may contain `null`s (they appear after removing some empty tokens during tokenizing). `ArrayView` is just wrapper for Array but it has fast methods (just one allocation) `take`, `drop`, `span` which are heavily used by `UriBuilder`.

Do you want to see this text as comment in code somewhere? Sorry, not quite understand your initial request.

adamw:
Thanks!

I meant in the code - PRs don't live too long, and on the other hand you often read code when you need to fix something. So if something isn't self-explanatory, a comment is useful. There's nothing wrong with the implementation, but mutable state often needs such comments :)

As for ArrayList vs ArrayBuffer - I tend to prefer to use scala collections, as it's a safer default when cross-building using native & js. I think here it works fine, but still as a rule staying in Scala-land is easier long-term (if possible)

danilbykov:
I added comments in code in several places and replaces `ArrayList` by `ArrayBuffer`. I see small performance degradation with `ArrayBuffer` (about 10%). But it is probably not so much.

PR #210, opened 2022-08-14
Add application/grpc media type

mbore:
required by https://github.com/softwaremill/tapir/pull/2345


adamw:
Sorry, binary incompatible ... we'll have to come up with a better encoding of enumerating media types so that we can evolve the interface.

But you should be able to define a custom media type without being it present in sttp-model?

mbore:
Yes, I am. I just wanted to add it to this model since it's a common media type, but will add it in tapir.

PR #214, opened 2022-08-31
Added map with status texts

Pask423:
Base for https://github.com/softwaremill/sttp/issues/1451. I would like to add set of status text here and use them in sttp core `Response`.


core/src/main/scala/sttp/model/StatusCode.scala
```diff
@@ -2,6 +2,7 @@ package sttp.model
 
 import internal.Validate._
 
+// TODO: When moving to v2 add status text as field
```
adamw, line 5: well this models a status code, why would it include a status text? :)

core/src/main/scala/sttp/model/StatusTexts.scala
```diff
@@ -0,0 +1,72 @@
...
+    GatewayTimeout -> "Gateway Timeout",
+    HttpVersionNotSupported -> "Http Version Not Supported",
+    VariantAlsoNegotiates -> "Variant Also Negotiates",
+    InsufficientStorage -> "Insufficient Storage",
+    LoopDetected -> "Loop Detected",
+    NotExtended -> "Not Extended",
+    NetworkAuthenticationRequired -> "Network Authentication Required"
+  )
+
+  def get(statusCode: StatusCode): String =
```
adamw, line 70: I think it would be better to return an `Option` ale let the user decide what to do if it's a non-standard status code.

Are these texts "default" descriptions, that is, might they differ in actual server responses? Why would need this map in the first place? :)

core/src/main/scala/sttp/model/StatusCode.scala
Pask423: As if I think correctly status text is to a degree part of status code as it should be same as name of status code https://developer.mozilla.org/en-US/docs/Web/API/Response/statusText

core/src/main/scala/sttp/model/StatusTexts.scala
Pask423: 1 OK 
2 Yes I intend to use it as default. In sttp core in `Response` class we allow to set status text as in the screen. I would like to write method that will use of these texts if sttp user does not provide one
image

core/src/main/scala/sttp/model/StatusCode.scala
adamw: To me it looks as if it's a different property on a `Response` class, which can be usually calculated from the status code, but definitely not part of it

core/src/main/scala/sttp/model/StatusTexts.scala
adamw: Ok, so maybe let's call this method `StatusTexts.default(StatusCode)`?

core/src/main/scala/sttp/model/StatusCode.scala
Pask423: ok, fixes incoming

core/src/main/scala/sttp/model/StatusTexts.scala
Pask423: Ok

adamw:
Thanks :)

PR #232, opened 2022-11-12
Added HttpVersion model

Pask423:


core/src/main/scala/sttp/model/HttpVersion.scala
```diff
@@ -0,0 +1,9 @@
+package sttp.model
+
+sealed trait HttpVersion
+object HttpVersion {
+  case object Default extends HttpVersion
```
adamw, line 5: I think that maybe as an enum, the `HttpVersion` shouldn't include `Default`. On the sttp-client side, we can use an optional. We might include `HTTP_3` for the future, though ;)

Pask423: done

adamw (approved):

PR #251, opened 2023-03-02
Added TooEarly StatusCode (425) to StatusCodes trait

domdorn:
TooEarly was missing
https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml


adamw:
Unfortunately adding a field to a trait isn't bin-compatible. Adding a `def`, though (w/ a comment that it's because of bin compat), should be fine. Not ideal, but better than nothing.

domdorn:
hmm.. i just did it the same way like the others.. should I change it to a def?

adamw:
Yes, but the others were part of the original release, and adding new fields is not possible due to binary compatibity requirements.

adamw:
Sigh ... scala 2.11 ... we'll be dropping 2.11 soon so I'll merge this PR then. Sorry :)

PR #270, opened 2023-06-18
Add value validation to header

wydra98:


core/src/main/scala/sttp/model/Header.scala
```diff
@@ -67,7 +67,21 @@ object Header {
   def unsafeApply(name: String, value: String): Header = safeApply(name, value).getOrThrow
 
   def safeApply(name: String, value: String): Either[String, Header] = {
-    Validate.all(validateToken("Header name", name))(apply(name, value))
+    Validate.all(validateToken("Header name", name), validateValue(value))(apply(name, value))
```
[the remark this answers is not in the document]
wydra98: These rules are from [here](https://www.rfc-editor.org/rfc/rfc9110.html#name-field-values). You mentioned this link in the [comment](https://github.com/softwaremill/sttp/issues/1823). I focused especially on the part: 


> A field value does not include leading or trailing whitespace. When a specific version of HTTP allows such whitespace to appear in a message, a field parsing implementation MUST exclude such whitespace prior to evaluating the field value.
> 
> Field values are usually constrained to the range of US-ASCII characters [[USASCII](https://www.rfc-editor.org/rfc/rfc9110.html#USASCII)]. Fields needing a greater range of characters can use an encoding, such as the one defined in [[RFC8187](https://www.rfc-editor.org/rfc/rfc9110.html#RFC8187)]. Historically, HTTP allowed field content with text in the ISO-8859-1 charset [[ISO-8859-1](https://www.rfc-editor.org/rfc/rfc9110.html#ISO-8859-1)], supporting other charsets only through use of [[RFC2047](https://www.rfc-editor.org/rfc/rfc9110.html#RFC2047)] encoding. Specifications for newly defined fields SHOULD limit their values to visible US-ASCII octets (VCHAR), SP, and HTAB. A recipient SHOULD treat other allowed octets in field content (i.e., [obs-text](https://www.rfc-editor.org/rfc/rfc9110.html#fields.values)) as opaque data.
>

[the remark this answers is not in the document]
adamw: Ah  ok :) Let's put it in a `Rfc9110` object just as the header name validation.

Though it's a should, not a must ... what is the current standard supported by browsers / servers? That is, what are the strict limitations on the charset?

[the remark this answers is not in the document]
adamw: That's what I found in Akka's docs: `// http://www.rfc-editor.org/errata_search.php?rfc=7230 errata id 4189`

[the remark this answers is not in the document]
wydra98: ok, I put this code in `Rfc9110` object. Do you mean which RFC version is current standard or something different?

core/src/main/scala/sttp/model/internal/Rfc9110.scala
```diff
@@ -0,0 +1,17 @@
+package sttp.model.internal
+
+object Rfc9110 {
+  private val availableWhitespaces = "\\x09\\x20"
+  private val VCHAR = "\\x21-\\x7E"
+  private val regex = s"^[$VCHAR]+([$availableWhitespaces]+[$VCHAR]+)*$$"
+
+  def validateValue(v: String): Option[String] = {
```
adamw, line 8: please add a link where the character set for header values is defined; is it called "value" in the spec, or something else? let's keep to the terminology used there

wydra98: In the spec is called "Field Value", so i updated the code

core/src/test/scalajvm/sttp/model/HeaderTests.scala
```diff
@@ -20,11 +20,18 @@ class HeaderTests extends AnyFlatSpec with Matchers {
 
   it should "validate status codes" in {
     Header.safeApply("Aut ho", "a bc") should matchPattern { case Left(_) => }
+    Header.safeApply(HeaderNames.Authorization, "Я ЛЮБЛЮ БОРЩ") should matchPattern { case Left(_) => }
+    Header.safeApply(HeaderNames.Authorization, " Я ЛЮБЛЮ БОРЩ") should matchPattern { case Left(_) => }
+    Header.safeApply(HeaderNames.Authorization, " xy z") should matchPattern { case Left(_) => }
+    Header.safeApply(HeaderNames.Authorization, "xy z\t") should matchPattern { case Left(_) => }
+    Header.safeApply(HeaderNames.Authorization, "xy \n z") should matchPattern { case Left(_) => }
     Header.safeApply(HeaderNames.Authorization, "xy z") should matchPattern { case Right(_) => }
+    Header.safeApply(HeaderNames.Authorization, "x \t y  z") should matchPattern { case Right(_) => }
```
adamw, line 29: is an empty string a valid header value?

core/src/main/scala/sttp/model/internal/Rfc9110.scala
```diff
@@ -0,0 +1,18 @@
+package sttp.model.internal
+
+//https://www.rfc-editor.org/rfc/rfc9110.html#name-field-values
+object Rfc9110 {
```
adamw, line 4: since this is internal, let's add `private[model]` like in the other (not all, sadly) internal packages

adamw (approved):
Once the two minor comments are resolved, let's release as 1.6 (next minor release). Thanks!

core/src/test/scalajvm/sttp/model/HeaderTests.scala
wydra98: Empty string is invalid. I based on quote: 

> Specifications for newly defined fields SHOULD limit their values to visible US-ASCII octets (VCHAR), SP, and HTAB.

but in your opinion, should it be considered valid?

adamw: well it's not a matter of opinion, but of specification ;) `SHOULD` is different than `MUST`, so let's find what the specification says about the hard requirements that are put on header values. But in the sentence quoted above, I don't see anything that would rule out an empty value.

wydra98: After discussion with @kciesielski and after reviewing this PR: https://github.com/softwaremill/sttp/pull/1855, we have come to the conclusion that it is best to consider an empty string as a valid value and not reject it as an exception.

PR #289, opened 2023-08-01
More efficient working with headers

plokhotnyuk:
Currently, it is quite expensive in terms of burned CPU and eaten memory bandwidth during parsing, validation
image
...and serialization
image

This PR reduces that overheads and improves working with headers.

### DIFF
--- a/core/src/main/scala/sttp/model/headers/AcceptEncoding.scala
+++ b/core/src/main/scala/sttp/model/headers/AcceptEncoding.scala
@@ -13,11 +13,11 @@ case class AcceptEncoding(encodings: List[WeightedEncoding]) {
 object AcceptEncoding {
 
   case class WeightedEncoding(encoding: String, weight: Option[BigDecimal]) {
-    override def toString: String = s"$encoding${weight.map(w => s";q=$w").getOrElse("")}"
+    override def toString: String = weight.fold(encoding)(w => s"$encoding;q=$w")
   }
 
   def parse(str: String): Either[String, AcceptEncoding] = {
-    val encodings = processString(str, List.empty)
+    val encodings = processString(str)
     if (encodings.isEmpty) Left(s"No encodings found in: $str")
     else {
       @tailrec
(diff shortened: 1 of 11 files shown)


plokhotnyuk:
@adamw Could, please review this PR?

kciesielski:
@plokhotnyuk Thank you! Performance improvements like this one are much appreciated, and we are definitely interested in subsequent PRs. Unfortunately, the core code has to compile for Scala 2.12 as well. Could you adjust your changes to this requirement?

plokhotnyuk:
@kciesielski Sorry, I missed to check that `sbt +core/test` doesn't run tests across Scala versions.

I've force pushed updated commit with fixes that had tested locally by the following sequence of commands:
```
sbt ++2.12.18 core/test
sbt ++2.13.11 core/test
sbt ++3.3.0 core/test
```

kciesielski:
@plokhotnyuk We use the `sbt-projectmatrix` plugin, which generates dynamic projects for each scala version. The `core` project is for Scala 2.13, `core2_12` for 2.12 and `core3` for Scala 3. The failed compilation happens on `core2_12/compile` (compile scope).

plokhotnyuk:
@kciesielski Thanks, for the heads up! Now `core2_12/compile` compiles without any errors.

kciesielski:
Awesome, thanks again!

core/src/main/scala/sttp/model/headers/AcceptEncoding.scala
```diff
@@ -34,40 +34,28 @@ object AcceptEncoding {
...
-    str.trim.split(",").toList match {
-      case x :: tail if x.nonEmpty =>
-        val range = parsSingleEncoding(x)
-        processString(tail.mkString(","), range :: acc)
-      case Nil => List(parsSingleEncoding(str))
-      case _   => acc
-    }
-  }
+  private def processString(str: String): List[WeightedEncoding] =
+    str.trim.split(",").map(x => parsSingleEncoding(x.trim)).reverse.toList // TODO: do we really need `.reverse` here?
```
adamw, line 38: regarding the TODO - yeah it looks weird, why would we need to reverse; if anything, we want to preserve the original ordering?

plokhotnyuk: Originally reversing was introduced by recursive version of previous implementation.

So my question in the TODO message was about need of reversion for the result list.

adamw: There was another tail-recursion with an accumulator which reversed the list - I moved the `reverse` to the more logical place: https://github.com/softwaremill/sttp-model/commit/ea43d9684b576c794f804c4bc200200b21d216de

Thanks for the PR, btw :)

PR #306, opened 2023-10-10
Add `otherParameters` to ContentTypeRange, update `MediaType#matches` to take into account `otherParameters`.

### LINKED ISSUE #305: `MediaType#matches` ignores parameters.
Originally posted in https://github.com/softwaremill/tapir/issues/3232.

kamilkloch:
Fixes #305.

### DIFF
--- a/core/src/main/scala/sttp/model/ContentTypeRange.scala
+++ b/core/src/main/scala/sttp/model/ContentTypeRange.scala
@@ -1,10 +1,10 @@
 package sttp.model
 
-import sttp.model.ContentTypeRange.{Wildcard, emptyParameters}
+import sttp.model.ContentTypeRange.{Wildcard, EmptyParameters}
 
 case class ContentTypeRange(mainType: String, subType: String, charset: String, otherParameters: Map[String, String]) {
   // required for binary compatibility
-  def this(mainType: String, subType: String, charset: String) = this(mainType, subType, charset, emptyParameters)
+  def this(mainType: String, subType: String, charset: String) = this(mainType, subType, charset, EmptyParameters)
 
   def copy(
       mainType: String = this.mainType,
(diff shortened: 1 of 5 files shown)


core/src/main/scala/sttp/model/ContentTypeRange.scala
```diff
@@ -2,13 +2,24 @@ package sttp.model
 
 import sttp.model.ContentTypeRange.Wildcard
 
-case class ContentTypeRange(mainType: String, subType: String, charset: String) {
+case class ContentTypeRange(mainType: String, subType: String, charset: String, otherParameters: Map[String, String]) {
```
adamw, line 5: this will for sure need some massaging to meet bin compat

kamilkloch: Add constructor for the former case?

adamw: Yes, that might work. Running mima locally should determine if there's anything more to do.

kamilkloch: It does not :/

adamw: `mimaReportBinaryIssues` doesn't work for you? Ah, maybe you need to fetch tags as described here: https://github.com/softwaremill/tapir#contributing

kamilkloch: Do you have any pattern on how to resolve the issue with `copy`?

adamw: Yes, you need to spell out the copy methods by hand, e.g.: https://github.com/softwaremill/tapir/blob/7cd79a1cf69cd2465841062abeb4b5ea988426c1/core/src/main/scala/sttp/tapir/EndpointIO.scala#L584

kamilkloch:
I checked tapir locally against the locally published sttp-model. What do you mean by submitting this as a PR? It will not compile (missing locally published sttp-model).

adamw:
I know, but some more changes are needed on the tapir side, no? :) I can then checkout both PRs locally & test

kamilkloch:
I did not need to alter a single line in tapir to make it compile, I beileve it should work as expected with the new sttp-model. I tried to add some tests, but got lost in the tapir testing machinery :/ (`ContentNegitiation`, `ServerContentNegotiationTests`). I would appreciate your help here. Thank you!

adamw:
`ServerContentNegotiationTests` is the right place to add a test :) If you add a test there, it will be ran when the tests for any specific interpreter's tests are run, e.g. `NettyFutureServerTest`.

The test itself consists of an endpoint (here you would need to create an endpoitn with the `oneOf`) and an sttp-client request to make (here you would need to include the appropriate header).

kamilkloch:
@adamw please take a look: https://github.com/softwaremill/tapir/pull/3243

adamw:
1.7.3 should be out soon

adamw:
Yeah maybe simply update the PR that is already there with the test?

PR #308, opened 2023-10-14
optimization: cache ClassTag[Token] to avoid thread blocking

fwbrasil:
## Problem

I'm working on a Tapir module for [Kyo](https://github.com/getkyo/kyo/tree/main/kyo-tapir/src/main/scala/kyo) and noticed that there's some thread blocking to materialize `ClassTag`s in `UriInterpolator`. It seems a mild problem that should affect mostly benchmarks but there could be pathological scenarios with more complex routing that make it more problematic:

sttp-baseline

## Solution

Cache the `ClassTag` in a local field to avoid the synchronized cache lookup. The change avoids the blocking in `UriInterpolator` but there are still other cases that arise from other collection operations:

sttp-optimized

## Notes

- As an alternative users could disable the `ClassTag` cache via `-Dscala.reflect.classtag.cache.disable=true` but it could have a performance penalty.
- If there's an interest, I could explore ways to avoid the remaining sources of blocking.

### DIFF
--- a/core/src/main/scala/sttp/model/UriInterpolator.scala
+++ b/core/src/main/scala/sttp/model/UriInterpolator.scala
@@ -5,6 +5,7 @@ import sttp.model.internal.{ArrayView, FastCharSet, FastCharMap, ParseUtils, Rfc
 import scala.annotation.tailrec
 import scala.collection.immutable.VectorBuilder
 import scala.collection.mutable.ArrayBuffer
+import scala.reflect.ClassTag
 
 trait UriInterpolator {
   implicit class UriContext(val sc: StringContext) {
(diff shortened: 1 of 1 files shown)


core/src/main/scala/sttp/model/UriInterpolator.scala
```diff
@@ -43,6 +44,8 @@ object UriInterpolator {
 
   private val startingUri = Uri(None, None, Uri.EmptyPath, Nil, None)
 
+  private implicit val tokenClassTag: ClassTag[Token] = ClassTag(classOf[Token])
```
adamw, line 47: Thanks! Maybe you could add a short comment, on why this value is present here - otherwise it might be puzzling for newcomers to the codebase

adamw:
> If there's an interest, I could explore ways to avoid the remaining sources of blocking.

Of course, we're always open to PRs, and performance improvements in particular :)

core/src/main/scala/sttp/model/UriInterpolator.scala
fwbrasil: Thanks for the quick review! I've just updated the change with a code comment.

fwbrasil:
> Of course, we're always open to PRs, and performance improvements in particular :)

Great! Do you have benchmarks you normally use to validate Tapir's performance?

adamw:
@fwbrasil we do have some very basic tests, comparing "vanilla" akkahttp/http4s servers vs the same servers, but with tapir added on top, to check the overhead that tapir imposes. However, running them is not part of CI or the release process: https://github.com/softwaremill/tapir/tree/master/perf-tests

PR #309, opened 2023-10-17
Fix matching content type range parameters in presence of wildcards

adamw:
Follow-up to #306

cc @kamilkloch - when the range is `*/*` or `sth/*`, we don't require parameters to match. Otherwise there are problems with boundaries in multiparts (and probably others)


kamilkloch:
Thank you! I missed that one :/ Also, I could not find an RFC for content negotiation with additional parameters, 
https://github.com/softwaremill/sttp-model/blob/d0c1856ef01547b9bc72ec63a2df0fa42ce6626d/core/src/main/scala/sttp/model/MediaType.scala#L29
is our sensible guess.

adamw:
yes we'll see if people report any further problems :)

PR #313, opened 2023-10-26
Only checking content type range parameters, if any are defined in the range. If so, they must match exactly with the media type.

adamw:
Follow-up to #309 and https://github.com/softwaremill/tapir/issues/3253

cc @kamilkloch - the current logic caused problems with multiparts; so I needed to change it once again. Currently, the `otherParameters` are **only** checked if they are defined on the range (excluding the `q` parameter). If the range has parameters, they must match exactly (ignoring case) with what's in the media type. 

This also requires fixing the new test you added in tapir: but the functionality can be retained by changing the ordering of cases in the `oneOf`, so that the wildcard ones (without parameters) are defined first. But this also means that you cannot define a catch-all case, which would handle all requests where `Accept`s ranges have arbitrary parameters. I think this is correct, though, as if you have a request which says it only accepts responses with given media type, and a parameter `x=y`, you then have to handle this explicitly in code.

### DIFF
--- a/core/src/main/scala/sttp/model/MediaType.scala
+++ b/core/src/main/scala/sttp/model/MediaType.scala
@@ -19,18 +19,27 @@ case class MediaType(
 
   def noCharset: MediaType = copy(charset = None)
 
-  // #2994 from tapir: when the media type doesn't define a charset, it shouldn't be taken into account in the matching logic
+  /** Checks if this media type matches the given content type range:
+    *
+    *   - the main and sub types must match, unless they are wildcards in the range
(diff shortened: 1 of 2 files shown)


kamilkloch:
@adamw Thank you for the heads-up. 

This way range `application/json; q=0.6` will not match media type `application/json; a=1`, which contradicts "Currently, the otherParameters are only checked if they are defined on the range (excluding the q parameter). "

kamilkloch:
I will update tapir code once we decide w.r.t. #314.

adamw:
That's already in progress in https://github.com/softwaremill/tapir/pull/3274 :)

PR #314, opened 2023-10-26
Ignore q parameter in range.

kamilkloch:
Follow-up to #313 and #309.


adamw:
Good catch, thanks! :)

PR #331, opened 2024-02-29
cache `MediaType.toString`

fwbrasil:
## Problem

On each request, Tapir uses the `toString` in `MediaType` to produce the header. In my [benchmark](https://github.com/fwbrasil/rinha-2024-q1), the instance of `MediaType` is always the same but the string has to be produced again on each request.

##  Solution

Make `toString` a lazy val so the value can be reused.

## Notes

- Are `MediaType` meant to be singletons? If yes, we could make `toString` a `val` to avoid the lazy val volatile read. It's likely a quite small overhead, though.

### DIFF
--- a/core/src/main/scala/sttp/model/MediaType.scala
+++ b/core/src/main/scala/sttp/model/MediaType.scala
@@ -63,7 +63,7 @@ case class MediaType(
 
   def isModel: Boolean = mainType.equalsIgnoreCase("model")
 
-  override def toString: String = {
+  override lazy val toString: String = {
     val sb = new java.lang.StringBuilder(32) // "application/json; charset=utf-8".length == 31 ;)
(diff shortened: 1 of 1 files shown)


adamw:
> Are MediaType meant to be singletons? If yes, we could make toString a val to avoid the lazy val volatile read. It's likely a quite small overhead, though.

There's a number of pre-defined instances, but during parsing in server interpreters, a new instance is created. We could do pooling of the instances (so that these are true singletons - I think akka-http does it), but I have no idea if that would have any significant value. If in your tests `lazy val` removes this bottleneck, let's go with that.

fwbrasil:
Thanks for the quick review! Yes, `lazy val` is fine with me. The overhead of the volatile read is very small, especially on x86 due to TSO.

PR #334, opened 2024-03-04
fix breaking change in MediaType.toString

fwbrasil:
## Problem

https://github.com/softwaremill/sttp-model/pull/331 changed `MediaType.toString` to become a `lazy val`, which is a breaking change. I'm seeing this issue while trying to publish a version of Tapir with the optimizations mentioned in https://github.com/softwaremill/tapir/issues/3552.

## Solution

Make the `lazy val` private and keep the same `toString()` interface as before.

### DIFF
--- a/core/src/main/scala/sttp/model/MediaType.scala
+++ b/core/src/main/scala/sttp/model/MediaType.scala
@@ -63,7 +63,9 @@ case class MediaType(
 
   def isModel: Boolean = mainType.equalsIgnoreCase("model")
 
-  override lazy val toString: String = {
+  // Cache 'toString' given that it's called in the hot path 
+  // of request processing to generate headers.
+  private lazy val toStringCache: String = {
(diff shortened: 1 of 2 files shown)


adamw:
Oh that's interesting ... what kind of errors do you see? Mima seems happy, but apparently there's more to compatibility ;)

core/src/main/scala/sttp/model/MediaType.scala
```diff
@@ -63,7 +63,7 @@ case class MediaType(
 
   def isModel: Boolean = mainType.equalsIgnoreCase("model")
 
-  override lazy val toString: String = {
+  private lazy val toStringCache: String = {
```
adamw, line 66: maybe you could add a comment explaining why this particular class has a `toStringCache` field?

fwbrasil: done

adamw:
thx :)

PR #336, opened 2024-03-15
Removed re-encoding of multipart content-disposition header values

### LINKED ISSUE #317: Browser compatibility rendering support for Content-Disposition header
Hi,

I have noticed currently Part model strictly follows the [RFC2183](https://www.rfc-editor.org/rfc/rfc2183) for `Content-Disposition` header. In our case we have a filename with non US-ASCII characters (UTF-8) and browsers render the filename weirdly due to pre encoding in the sttp Part model. 

```
  def contentDispositionHeaderValue: String = {
    def encode(s: String): String = new String(s.getBytes("utf-8"), "iso-8859-1")
    "form-data; " + dispositionParamsSeq.map { case (k, v) => s"""$k="${encode(v)}"""" }.mkString("; ")
  }

```
Is there a workaround to overcome this issue or have a plan to implement the feasibility to switch modes while sending multiparts?  
apache provide a similar feature when building multiparts as `STRICT`,`BROWSER_COMPATIBLE`,`RFC6532`

ghik:
Closes #317 (with additional changes required in `sttp`)

### DIFF
--- a/core/src/main/scala/sttp/model/Part.scala
+++ b/core/src/main/scala/sttp/model/Part.scala
@@ -41,10 +41,8 @@ case class Part[+T](
   // enumerate all variants so that overload resolution works correctly
   override def header(h: String): Option[String] = super.header(h)
 
-  def contentDispositionHeaderValue: String = {
-    def encode(s: String): String = new String(s.getBytes("utf-8"), "iso-8859-1")
-    "form-data; " + dispositionParamsSeq.map { case (k, v) => s"""$k="${encode(v)}"""" }.mkString("; ")
-  }
+  def contentDispositionHeaderValue: String =
+    "form-data; " + dispositionParamsSeq.map { case (k, v) => s"""$k="$v"""" }.mkString("; ")
 
   def dispositionParams: Map[String, String] = dispositionParamsSeq.toMap
 
--- a/core/src/main/scala/sttp/model/Part.scala
+++ b/core/src/main/scala/sttp/model/Part.scala
@@ -41,8 +41,31 @@ case class Part[+T](
   // enumerate all variants so that overload resolution works correctly
   override def header(h: String): Option[String] = super.header(h)
 
-  def contentDispositionHeaderValue: String =
-    "form-data; " + dispositionParamsSeq.map { case (k, v) => s"""$k="$v"""" }.mkString("; ")
+  /**
+   * Returns the value of the `Content-Disposition` header, which should be used when sending this part in a
+   * multipart request.
+   *
+   * The syntax is specified by [[https://datatracker.ietf.org/doc/html/rfc6266#section-4.1 RFC6266 section 4.1]].
+   * For safety and simplicity, disposition parameter values are represented as `quoted-string`, defined in
+   * [[https://datatracker.ietf.org/doc/html/rfc9110#section-5.6.4 RFC9110 section 5.6.4]].
+   *
+   * `quoted-string` allows usage of visible ASCII characters (`%x21-7E`), except for `"` and `\`, which must be escaped
+   * with a backslash. Additionally, space and horizontal tab is allowed, as well as octets `0x80-FF` (`obs-data`).
+   * In practice this means that - while not explicitly allowed - non-ASCII UTF-8 characters are valid
+   * according to this grammar. Additionally, [[https://datatracker.ietf.org/doc/html/rfc6532#section-3.2 RFC6532]]
+   * makes it more explicit that non-ASCII UTF-8 is allowed. Control characters are not allowed.
+   *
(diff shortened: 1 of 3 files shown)


core/src/main/scala/sttp/model/Part.scala
```diff
@@ -41,10 +41,8 @@ case class Part[+T](
   // enumerate all variants so that overload resolution works correctly
   override def header(h: String): Option[String] = super.header(h)
 
-  def contentDispositionHeaderValue: String = {
-    def encode(s: String): String = new String(s.getBytes("utf-8"), "iso-8859-1")
-    "form-data; " + dispositionParamsSeq.map { case (k, v) => s"""$k="${encode(v)}"""" }.mkString("; ")
-  }
+  def contentDispositionHeaderValue: String =
+    "form-data; " + dispositionParamsSeq.map { case (k, v) => s"""$k="$v"""" }.mkString("; ")
```
adamw, line 45: can we somehow test the "correct" behavior?

btw. - I assume file names can't contain `"` or `;` - are they specified to be escaped somehow by the RFC?

ghik: Can you clarify what you mean by "correct" behaviour?

adamw: well ... before we encoded certain header values incorrectly. So there should exist a test which wasn't passing before, and is passing now :)

ghik: Yeah, there was no test (in fact, no usage at all) of this method in `sttp-model`. Sure, I'll add one.

As for the treatment of special characters - I'm a bit lost with the multitude of RFCs and different encodings, but I'll try to dig through them. So far I assumed that since the previous implementation didn't care about this, I can probably ignore that too just to fix the immediate problem (adding more strict validation and encoding could be solved in another issue).

adamw: Yeah RFC are confusing ;) I think the original impl came from the fact that I read on some RFC that multipart header values should be ISO-encoded. But the living standard might be different, or there might be a new RFC. If you find anything, please add it as a comment in code, so that we know where the behavior comes from.

ghik: Added escaping for `"` and `\` required by RFC and an explanation with references to grammars etc.

core/src/main/scala/sttp/model/Part.scala
```diff
@@ -41,9 +41,30 @@ case class Part[+T](
...
+   *
+   * `quoted-string` allows usage of visible ASCII characters (`%x21-7E`), except for `"` and `\`, which must be escaped
+   * with a backslash. Additionally, space and horizontal tab is allowed, as well as octets `0x80-FF` (`obs-data`).
+   * In practice this means that - while not explicitly allowed - non-ASCII UTF-8 characters are valid
+   * according to this grammar. Additionally, [[https://datatracker.ietf.org/doc/html/rfc6532#section-3.2 RFC6532]]
+   * makes it more explicit that non-ASCII UTF-8 is allowed. Control characters are not allowed.
+   *
+   * This method makes sure that `"` and `\` are escaped, while leaving possible rejection of forbidden characters to
+   * lower layers (`sttp` backends).
+   */
```
adamw, line 60: ❤️ all clear, thanks! :)

core/src/main/scala/sttp/model/Part.scala
```diff
@@ -41,9 +41,30 @@ case class Part[+T](
...
   def contentDispositionHeaderValue: String = {
-    def encode(s: String): String = new String(s.getBytes("utf-8"), "iso-8859-1")
-    "form-data; " + dispositionParamsSeq.map { case (k, v) => s"""$k="${encode(v)}"""" }.mkString("; ")
+    def escape(str: String): String = str.flatMap {
+      case '"' => "\\\""
+      case '\\' => "\\\\"
+      case c => c.toString
+    }
```
adamw, line 66: won't a .replace be faster? maybe I'm biased after @kciesielski 's performance tests, but this will convert the string to a collection I suppose, and then back to a string?

adamw (approved):

core/src/main/scala/sttp/model/Part.scala
ghik: At least in Scala 2.13, there's a specialized variant of `String.flatMap` which uses a `StringBuilder` directly.
OTOH, `replace` would have to be called twice (unless using `replaceAll` but this one uses a regex).

So, superficially, `flatMap` seems a bit safer performance-wise, although if you have some benchmark results which suggest otherwise then I'm happy to use a proven, faster implementation. If we _really_ want to optimize it then we should probably avoid any escaping when there's nothing to escape. But then everything ends up in a string interpolation anyway so we would probably need to rewrite this entire method to use a single `StringBuilder`.

PR #344, opened 2024-05-04
Add one more delimiter sign to Rfc3986

KarinaTyulebaeva:
Hello! Surprisingly it turned out that we need one additional delimiter character to correctly encode urls


adamw:
The `|` is not listed here: https://datatracker.ietf.org/doc/html/rfc3986#section-2.2

Can you maybe share a use-case of how did you encounter this problem?

KarinaTyulebaeva:
> The `|` is not listed here: https://datatracker.ietf.org/doc/html/rfc3986#section-2.2
> 
> Can you maybe share a use-case of how did you encounter this problem?

Ah, sorry, misread the rfc, I thought there was such a sign there

PR #350, opened 2024-05-22
Added method for parsing string body

Pask423:


core/src/main/scalajs/sttp/model/internal/UriCompatibility.scala
```diff
@@ -11,4 +11,6 @@ private[sttp] object UriCompatibility {
   }
 
   def encodeQuery(s: String, enc: String): String = URIUtils.encodeURIComponent(s)
+
+  def encodeBodyPart(s: String, enc: String): String = URIUtils.encodeURIComponent(s)
```
adamw, line 15: That's for encoding a part of a multipart body?

core/src/main/scalajvm/sttp/model/internal/UriCompatibility.scala
```diff
@@ -18,5 +16,7 @@ private[sttp] object UriCompatibility {
     Rfc3986.encode(Rfc3986.Host)(noSpecialChars)
   }
 
-  def encodeQuery(s: String, enc: String): String = URLEncoder.encode(s, enc)
+  def encodeQuery(s: String, enc: String): String = Rfc3986.encode(s, enc, Rfc3986.Unreserved)
```
adamw, line 19: do we also want to change how queries are encoded?

core/src/main/scala/sttp/model/Uri.scala
```diff
@@ -709,7 +709,7 @@ object Uri extends UriInterpolator {
 
   object QuerySegmentEncoding {
 
-    /** Encodes all reserved characters using [[java.net.URLEncoder.encode()]]. */
+    /** Encodes all reserved characters using [[sttp.model.internal.Rfc3986]] with characters from [[sttp.model.internal.Rfc3986.Unreserved]].  */
```
adamw, line 712: I think the `All` encoding is specifically created to encode *all* characters, as oppose to `Standard`, which only encodes the ones mentioned in the specs. So I wouldn't change this.

core/src/main/scala/sttp/model/internal/Rfc3986.scala
```diff
@@ -15,17 +15,36 @@ object Rfc3986 {
...
     * @param encodePlus
     *   Should `+` (which is the encoded form of space in the query) be %-encoded.
     */
   def encode(allowedCharacters: Set[Char], spaceAsPlus: Boolean = false, encodePlus: Boolean = false)(
       s: String
+  ): String =
+    encode(s, "utf-8", allowedCharacters, spaceAsPlus, encodePlus)
+
+
+  def encode(s: String, enc: String, allowedCharacters: Set[Char]): String =
```
adamw, line 30: missing comment

core/src/main/scala/sttp/model/Uri.scala
Pask423: Previously this was encoded by URLEncoder so in fact only change now is that `~`  won't be encoded

core/src/main/scalajs/sttp/model/internal/UriCompatibility.scala
Pask423: Yes, but not sure if we have multipart support for JS anywhere, I added it for sake of compatibility with other implementations of `UriCompatibility`

adamw: Yeah sounds good, ok

core/src/main/scala/sttp/model/Uri.scala
adamw: I this change required to fix the bug in sttp? It feels there's a couple of changes done at one time

Pask423: I thought it will be a good idea to have same encoding here and for multiparts, I can revert it

adamw: I'd keep the changes minimal just to address the sttp bug. We can discuss other changes, but in separate PRs. That way we can better keep track of things that might get broken by the changes.

core/src/main/scala/sttp/model/internal/Rfc3986.scala
```diff
@@ -15,17 +15,44 @@ object Rfc3986 {
...
+      s: String,
+      enc: String,
+      allowedCharacters: Set[Char],
+      spaceAsPlus: Boolean,
+      encodePlus: Boolean
   ): String = {
     val sb = new StringBuilder()
     // based on https://gist.github.com/teigen/5865923
-    for (b <- s.getBytes("UTF-8")) {
+    for (b <- s.getBytes(enc)) {
```
plokhotnyuk, line 55: Please use `while` loop to avoid boxing/unboxing of byte values

core/src/test/scalajs/sttp/model/UriTestsHelper.scala
```diff
@@ -0,0 +1,25 @@
+package sttp.model
+
+import org.scalatest.TryValues
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers
+import sttp.model.Uri._
+
+class UriTestsHelper extends AnyFunSuite with Matchers with TryValues {
```
adamw, line 8: what are we testing here, and why is this different from `UriTests`?

Pask423: JS code seem to treat `~` as unreserved character while JVM one is encoding it so this case requited platform specific approach.

adamw: so these are platform-specific tests? the name doesn't suggest this at all

at a minimum, we need clear comments on why these tests are separate; but better (or additionally), we should rename the class. I think in other instances, we have a `UriTestsExtensions` trait, which is extended by `UriTests`, and is empty for unsupported platforms (see e.g. in tapir). So maybe let's do the same here

Pask423: ok I will go with `UriTestsExtensions`

PR #383, opened 2025-02-17
github actions workflows refactor

Blef666:


.github/workflows/scala-steward.yml
```diff
@@ -7,22 +7,9 @@ on:
...
-          cache: 'sbt'
-      - name: Launch Scala Steward
-        uses: scala-steward-org/scala-steward-action@v2
-        with:
-          author-name: scala-steward
-          author-email: scala-steward
-          github-token: ${{ secrets.REPO_GITHUB_TOKEN }}
-          repo-config: .scala-steward.conf
-          ignore-opts-files: false
+  call-workflow-scala-steward:
```
adamw, line 10: can't we just call the job `scala-steward`?

.github/workflows/scala-steward.yml
```diff
@@ -7,22 +7,9 @@ on:
...
+    uses: softwaremill/github-actions-workflows/.github/workflows/scala-steward.yml@main
+    with:
+      java-version: '11'
```
adamw, line 13: can we just have a default java version 11?

.github/workflows/scala-steward.yml
```diff
@@ -7,22 +7,9 @@ on:
...
-      - name: Launch Scala Steward
-        uses: scala-steward-org/scala-steward-action@v2
-        with:
-          author-name: scala-steward
-          author-email: scala-steward
-          github-token: ${{ secrets.REPO_GITHUB_TOKEN }}
-          repo-config: .scala-steward.conf
-          ignore-opts-files: false
+  call-workflow-scala-steward:
+    uses: softwaremill/github-actions-workflows/.github/workflows/scala-steward.yml@main
```
adamw, line 11: other actions use syntax like `scala-steward-org/scala-steward-action@v2`, this is much longer and has the `.github/worksflows` - why?

.github/workflows/scala-steward.yml
Blef666: Yes, we can, although I based it on the [documentation](https://docs.github.com/en/actions/sharing-automations/reusing-workflows) and all of the caller jobs in there are named `call-workflow-*` so I assumed that it's a convention. Maybe this assumption was incorrect. Do we want to change it?

.github/workflows/scala-steward.yml
Blef666: Yes, we can. 60% of our workflows use 11, 40% use 17 and 21. Should I change it?

.github/workflows/scala-steward.yml
Blef666: It's because one is an Action and the other one is a Workflow. These are two different things. An Action is a self contained piece of code written e.g. in JS and Workflow is not a code, but a collection of correlated Actions contained in yaml files. Workflows have to be in the `.github/worksflows` path, otherwise Github Actions won't be able to use them.

.github/workflows/scala-steward.yml
adamw: let's just use 21 for this

Blef666: 21? You just wanted 11. If we want to keep latest LTS version as default, then we will change it in the future and this will force us to change it also in other repositories using these workflows. Personally I'd rather have no default version so that everyone used it consciously, without mistakes.

adamw: we should use 11 when we build jar-s (artifacts) that are published. For this job, where it's just running some java code, we can use latest LTS

.github/workflows/scala-steward.yml
adamw: Can we establish if it's a convention or just an example in docs? These names then show up in the CI builds, so it would be nice to have them relatively clean. But if it's a convention, let's stick to that.

Also, the README for the workflows project could benefit from a "architecture" section, linking to the appropriate GH Actions docs / convention documents to understand why things are done the way they are.

.github/workflows/scala-steward.yml
adamw: Ok, let's add this to the readme as well :)

.github/workflows/scala-steward.yml
Blef666: I'm a little confused. Are you suggesting that the default version 11 should be only in publish workflow, and other workflows using scala (build-scala, mima, scala-steward) should have default 21?

.github/workflows/scala-steward.yml
adamw: Yes, I think we should keep the names as simple and as meaningful as possible

.github/workflows/scala-steward.yml
adamw: we can just use 11 everywhere, scala-steward's version doesn't matter that much

.github/workflows/scala-steward.yml
Blef666: done

.github/workflows/scala-steward.yml
Blef666: done

.github/workflows/ci.yml
```diff
@@ -6,175 +6,46 @@ on:
     branches: ['**']
     tags: [v*]
 jobs:
-  ci:
+  build-scala:
```
adamw, line 9: can we just call it `build` or `ci` as previously? there are no other `build` steps, so the `-scala` isn't really meaningful ;)

.github/workflows/ci.yml
```diff
@@ -6,175 +6,46 @@ on:
...
-          find $HOME/.ivy2/cache                       -name "*-LM-SNAPSHOT*"       -delete || true
-          find $HOME/.cache/coursier/v1                -name "ivydata-*.properties" -delete || true
-          find $HOME/.sbt                              -name "*.lock"               -delete || true
+    with:
+      java-opts: "-Xmx4G"
 
-  publish:
-    name: Publish release
-    needs: [ci]
+  publish-release:
```
adamw, line 33: similarly, lets simply use `publish` or `release` here

.github/workflows/ci.yml
Blef666: job names should be [verbs](https://nimblehq.co/compass/development/code-conventions/github-actions/#convention-4) so I'll stay with `build`

adamw:
Great, so ... we can merge? :) Or anything left to do? :)

Blef666:
I think we can.

Blef666:
To be honest I haven't considered this option. It appears that it should be [possible](https://github.com/orgs/community/discussions/63863#discussioncomment-6736608), but I haven't checked that.

PR #394, opened 2025-03-24
`Uri.toStringSafe` for hiding sensitive query params

aartigao:
This is the first part that needs to be implemented for adding the feature described in [sttp#2497](https://github.com/softwaremill/sttp/issues/2497).


adamw:
Thanks, release coming up :)

PR #408, opened 2025-07-02
Add support for filename* on disposition header getter

mayman:
We came across an issue where when creating a multipart, if the name contains non ascii characters, they get added as `filename*` in the disposition header.

This PR is to address the issue, so we can retrieve the non-ascii file name correctly.


adamw:
But either way, this correctly implements the RFC, hence merging :)

PR #465, opened 2026-07-10
Add RFC 10008 (The HTTP QUERY Method) support

desiderantes:
This adds basic support so it can be used to build proper RFC 10008 support in consumers.


adamw:
Thanks! (added a small MiMA tweak)

PR #478, opened 2026-08-28
Add support for SSE comments

### LINKED ISSUE #379: `ServerSentEvent` type does not match the WhatWG specification WRT comments
The model defined in https://github.com/softwaremill/sttp-model/blob/afd84a6e357c53b79ee7344a890223758285fbc5/core/src/main/scala/sttp/model/sse/ServerSentEvent.scala#L5-L18 does not match the specification at https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation

Specifically, this part:

> If the line starts with a U+003A COLON character (:)
> 
>     Ignore the line.

Therefore there is currently no way to send or receive "comment" messages, which are useful to perform keep-alives.

One could argue that you can just send `event: keep-alive` here, to which I agree, but just noticed this discrepancy and wondered whether we should add a way to send out these comment messages.

magdzikk:
Fixes #379.

`ServerSentEvent` gains a `comments: List[String]` field, so comment lines (those starting with `:`) round-trip through both `parse` and `toString`. `ServerSentEvent.comment("ping")` builds the keep-alive frame the [WhatWG specification](https://html.spec.whatwg.org/multipage/server-sent-events.html) recommends sending every 15 seconds or so, to stop legacy proxies from dropping an idle connection.

### Design notes

`List[String]` rather than `Option[String]`, because a single event block may carry several comment lines.

There is deliberately **no assertion** that a comment and data aren't both set (as considered in the issue discussion). The specification ignores comments per *line*, not per event, so a block such as

```
: keep-alive
data: abc
```

is **valid** and still dispatches `abc`. A guard would reject valid streams.

### Binary compatibility

Preserved with the same approach already used for `ContentTypeRange`: the old-arity constructor, `copy` and `apply` are kept alongside the new ones, and `mimaReportBinaryIssues` is green.

MiMa does not see everything, though. The source-level and runtime-shape changes it cannot report are listed under Migration below.

### Migration

Everything here follows from `ServerSentEvent` gaining a field. Recompiling against the new version is the general fix; the two behaviour changes need a code change as well.

**Comment-only events are no longer empty.** A `: ping` block used to parse to `ServerSentEvent()`, and now parses to an event with `comments` set, so code that dropped keep-alives by comparing against an empty event stops dropping them:

```scala
events.filterNot(_ == ServerSentEvent())  // no longer drops keep-alives
events.filterNot(_.hasNoFields)           // use this instead
```

`hasNoFields` is true when an event carries no data, event type, id or retry, so it drops exactly the set that the old comparison dropped — keep-alives, empty blocks, and blocks of unknown fields alike.

**Events parsed from a stream containing comments no longer equal hand-built events.** Streams often open with a `:` handshake comment, so expected values need updating, or the comments dropped before comparing:

```scala
ServerSentEvent.parse(lines) == ServerSentEvent(Some("x"))                      // false once a comment is present
ServerSentEvent.parse(lines).copy(comments = Nil) == ServerSentEvent(Some("x")) // compare without them
```

**Patterns over the four original fields no longer compile.** Add a fifth `_`, or use the accessors:

```scala
case ServerSentEvent(d, e, i, r)    => ... // does not compile
case ServerSentEvent(d, e, i, r, _) => ...
```

**Code compiled against an earlier version has to be recompiled.** The changes below leave the erased signatures intact, so such code links and then fails at runtime, and MiMa cannot report it:

- `ServerSentEvent.unapply` now returns a `Tuple5`. A caller compiled earlier fails with `ClassCastException: scala.Tuple5 cannot be cast to scala.Tuple4`. Compiled *patterns* are unaffected — they lower to accessor calls rather than to `unapply`.
- Scala 3 codecs derived through `Mirror` read a fifth element. One compiled earlier fails with `IndexOutOfBoundsException: 4`.

**Serialised forms change.**

- `serialVersionUID` changes from `5877847238202557287` to `-2395116303002558390`, so Java-serialised events do not cross the version boundary.
- JSON codecs derived from the case class gain a `comments` field. circe and play-json ignore Scala default arguments, so their derived decoders reject JSON that lacks it; jsoniter-scala and zio-json read the default and are unaffected.


core/src/main/scala/sttp/model/sse/ServerSentEvent.scala
```diff
@@ -6,22 +6,58 @@ case class ServerSentEvent(
...
+  // required for binary compatibility
+  def copy(
+      data: Option[String],
+      eventType: Option[String],
+      id: Option[String],
+      retry: Option[Int]
+  ): ServerSentEvent = ServerSentEvent(data, eventType, id, retry, this.comments)
+
   override def toString: String = {
+    val _comments: Array[Option[String]] = comments.map(comment => Some(s": $comment")).toArray
```
adamw: Comments are not split on `\n`, unlike data one line below. `comment("a\nb").toString` emits a bare `b` line, which is dropped on re-parse. `comment("ping\n\n")` emits a blank line, which ends the event. Since `comment(...)` accepts arbitrary text, a `\n` can also inject other fields (`comment("x\ndata: y")`).

Suggestion (also drops the `Some`/`Array` wrapping):
```scala
val _comments = comments.flatMap(_.split("\n")).map(c => Some(s": $c")).toArray
```

core/src/main/scala/sttp/model/sse/ServerSentEvent.scala
```diff
@@ -6,22 +6,58 @@ case class ServerSentEvent(
...
+  /** An event consisting of a single comment line. Such events are ignored by clients, and can be used to keep the
+    * connection alive, so that it isn't dropped by proxies.
+    */
+  def comment(comment: String): ServerSentEvent = ServerSentEvent(comments = List(comment))
+
   // https://html.spec.whatwg.org/multipage/server-sent-events.html
   def parse(event: List[String]): ServerSentEvent = {
     event.foldLeft(ServerSentEvent()) { (event, line) =>
-      if (line.startsWith("data:")) combineData(event, removeLeadingSpace(line.substring(5)))
+      if (line.startsWith(":")) event.copy(comments = event.comments :+ removeLeadingSpace(line.substring(1)))
```
adamw: Behavior change worth a release note: comment-only blocks used to parse to `ServerSentEvent()`. Code that skips keep-alives via `_ != ServerSentEvent()` or `case ServerSentEvent(None, None, None, None)` will now pass them through.

It would help to add a migration target, e.g.:
```scala
/** True if the event carries no data, event type, id or retry - only comments, if any.
  * Clients ignore comments, so such events (e.g. keep-alive pings) can usually be skipped.
  */
def isCommentOnly: Boolean = data.isEmpty && eventType.isEmpty && id.isEmpty && retry.isEmpty
```

Follow-up: audit tapir & sttp client for `ServerSentEvent()` comparisons/patterns and update them once this is released.

core/src/main/scala/sttp/model/sse/ServerSentEvent.scala
```diff
@@ -35,8 +71,8 @@ object ServerSentEvent {
 
   private def combineData(event: ServerSentEvent, newData: String): ServerSentEvent = {
     event match {
-      case e @ ServerSentEvent(Some(oldData), _, _, _) => e.copy(data = Some(s"$oldData\n$newData"))
-      case e @ ServerSentEvent(None, _, _, _)          => e.copy(data = Some(newData))
+      case e @ ServerSentEvent(Some(oldData), _, _, _, _) => e.copy(data = Some(s"$oldData\n$newData"))
```
adamw: These patterns only needed changing because they depend on the field count. An accessor form avoids the same edit next time a field is added:
```scala
event.copy(data = Some(event.data.fold(newData)(old => s"$old\n$newData")))
```

adamw (commented):
Automated review by Claude (posted by @adamw). Main finding: comments are not split on newlines in `toString`. Details inline.

core/src/test/scala/sttp/model/sse/ServerSentEventTest.scala
```diff
@@ -5,7 +5,7 @@ import org.scalatest.matchers.should.Matchers
 
 class ServerSentEventTest extends AnyFlatSpec with Matchers {
   val data = List(
-    (List(": this is a test stream"), ServerSentEvent()),
+    (List(": this is a test stream"), ServerSentEvent(comments = List("this is a test stream"))),
```
adamw, line 8: This was the only row reaching the final `else` in `parse` (ignore unknown lines). To keep that covered, add e.g. `(List("foo: bar", "data: x"), ServerSentEvent(Some("x")))`.

core/src/test/scala/sttp/model/sse/ServerSentEventTest.scala
```diff
@@ -24,7 +24,17 @@ class ServerSentEventTest extends AnyFlatSpec with Matchers {
     (
       List("data: event1 data", "event: event1", "id: id1", "retry: 5"),
       ServerSentEvent(Some("event1 data"), Some("event1"), Some("id1"), Some(5))
-    )
+    ),
+    (
+      List(": keep-alive", "data: with a comment"),
```
adamw: This row is covered by the next one, which also tests a comment after data. Could be dropped.

core/src/main/scala/sttp/model/sse/ServerSentEvent.scala
magdzikk: Fixed, and I'll take care about the follow-up check

magdzikk: @adamw I checked usages in tapir and sttp client and it seems they're unaffected. In these libraries, the filtering only ever occurs on line groups, never on parsed events, so no action is required there after we merge this change.

core/src/main/scala/sttp/model/sse/ServerSentEvent.scala
```diff
@@ -6,22 +6,63 @@ case class ServerSentEvent(
...
+      retry: Option[Int]
+  ): ServerSentEvent = ServerSentEvent(data, eventType, id, retry, this.comments)
+
+  /** True if the event carries no data, event type, id or retry - only comments, if any. Clients ignore comments, so
+    * such events (e.g. keep-alive pings) can usually be skipped.
+    */
+  def isCommentOnly: Boolean = data.isEmpty && eventType.isEmpty && id.isEmpty && retry.isEmpty
+
   override def toString: String = {
+    val _comments = comments.flatMap(_.split("\n")).map(comment => Some(s": $comment")).toArray
```
adamw: Comments are split only on `\n`, but the SSE spec also treats `\r` and `\r\n` as line terminators. So `comment("x\rdata: y").toString` emits the single line `: x\rdata: y`, which a browser `EventSource` parses as a comment plus a real `data: y` field — the injection the new test aims to prevent. Suggestion: split on `\r\n|\r|\n`.

_(generated by Claude)_

adamw: `split` drops trailing empty tokens, so `comment("\n").toString` is `""` — a keep-alive built from such a comment emits nothing. It's also inconsistent: `comment("")` emits `: `. Normalizing comments at construction time would fix this, and would also make multi-line comments round-trip (`comment("a\nb")` currently re-parses as `List("a", "b")`).

_(generated by Claude)_

magdzikk: @adamw Fixed, but a similar injection can happen in other fields - I added a separate issue for that https://github.com/softwaremill/sttp-model/issues/480 and a PR https://github.com/softwaremill/sttp-model/pull/481

core/src/main/scala/sttp/model/sse/ServerSentEvent.scala
```diff
@@ -6,22 +6,74 @@ case class ServerSentEvent(
...
 object ServerSentEvent {
+  private val LineTerminators = "\r\n|\r|\n"
+
+  def apply(
+      data: Option[String] = None,
+      eventType: Option[String] = None,
+      id: Option[String] = None,
+      retry: Option[Int] = None,
+      comments: List[String] = Nil
+  ): ServerSentEvent = new ServerSentEvent(data, eventType, id, retry, comments.flatMap(_.split(LineTerminators)))
```
adamw: Should the invariant be that `comments` never contains line terminators? Right now only this `apply` enforces it: `new ServerSentEvent(...)` — and Scala 3 codecs derived via `Mirror`, which construct through `fromProduct` — can still create `comments = List("a\nb")`. `toString` output stays safe either way, but such an instance doesn't round-trip (`parse` gives the split form) and isn't equal to the same event built via `apply`.

So: do we enforce the invariant everywhere, or allow both forms?

Also, maybe we should have property-based tests for the parse/serialise round-trip — they'd catch cases like this.

core/src/main/scala/sttp/model/sse/ServerSentEvent.scala
```diff
@@ -6,22 +6,74 @@ case class ServerSentEvent(
...
+      data: Option[String],
+      eventType: Option[String],
+      id: Option[String],
+      retry: Option[Int]
+  ): ServerSentEvent = new ServerSentEvent(data, eventType, id, retry, Nil)
+
+  /** An event consisting of a single comment line. Such events are ignored by clients, and can be used to keep the
+    * connection alive, so that it isn't dropped by proxies.
+    */
+  def comment(comment: String): ServerSentEvent = ServerSentEvent(comments = List(comment))
```
adamw: A comment that is only line terminators disappears entirely: `"\n".split("\r\n|\r|\n")` is an empty array, so `comment("\n")` equals `ServerSentEvent()` and serializes to `""` — a keep-alive built from such a string sends nothing. Meanwhile `comment("")` gives `": "`. Worth a test pinning what should happen here.

core/src/main/scala/sttp/model/sse/ServerSentEvent.scala
```diff
@@ -6,22 +6,74 @@ case class ServerSentEvent(
...
+
   // https://html.spec.whatwg.org/multipage/server-sent-events.html
   def parse(event: List[String]): ServerSentEvent = {
     event.foldLeft(ServerSentEvent()) { (event, line) =>
-      if (line.startsWith("data:")) combineData(event, removeLeadingSpace(line.substring(5)))
+      if (line.startsWith(":")) event.copy(comments = event.comments :+ removeLeadingSpace(line.substring(1)))
```
adamw: Every branch here calls `copy`, which goes through the 5-arg `apply` — so the whole accumulated comments list is re-split on every parsed line. That's always a no-op (parse input is already split into lines), and with the `:+` append it's quadratic in comment count per event. A `copy` that forwards to the constructor would avoid it.

core/src/main/scala/sttp/model/sse/ServerSentEvent.scala
```diff
@@ -6,22 +6,74 @@ case class ServerSentEvent(
...
     val _event = eventType.map(event => s"event: $event")
     val _id = id.map(id => s"id: $id")
     val _retry = retry.map(retryCount => s"retry: $retryCount")
-    (_data :+ _event :+ _id :+ _retry).flatten.mkString("\n")
+    ((_comments ++ _data) :+ _event :+ _id :+ _retry).flatten.mkString("\n")
   }
 }
 
 object ServerSentEvent {
+  private val LineTerminators = "\r\n|\r|\n"
```
adamw, line 54: `String.split` with a multi-char pattern recompiles the regex on every call (master's `split("\n")` hit the single-char fast path). Since this runs per comment in `apply` and again in `toString`, consider precompiling: `private val LineTerminators = java.util.regex.Pattern.compile("\r\n|\r|\n")`, then `LineTerminators.split(s)`.

magdzikk: I think we could solve it the same way as in https://github.com/softwaremill/sttp-model/commit/ab31bfe701c88f485b39b3b768b7c1037b2a7da3 - if https://github.com/softwaremill/sttp-model/pull/481 gets merged first, we can rebase this PR onto it, and use it here

core/src/main/scala/sttp/model/sse/ServerSentEvent.scala
magdzikk: Ah, I missed `fromProduct`. In this case I think allowing both forms is better. `comments` would then be a plain record of what was passed, and `toString` would keep splitting when writing.

Enforcing it everywhere would make the model throw, which doesn't fit `safeApply`/`unsafeApply` here, and would turn a newline in a comment into an exception inside derived decoders. 

So I'd remove the split from `apply` and keep it in comment(...)`, so the factory still hands back one entry per line. 

That would also remove the [re-splitting](https://github.com/softwaremill/sttp-model/pull/478#discussion_r3905855702) in parse.

On property tests: currently there's no scalacheck in the dependencies, so we can consider just adding a table-driven round-trip test over the special shapes ("", "\n", "\r", "\r\n", "a\nb", "ping\n\n", multi-item lists). But happy to add property tests if you don't mind adding that dependency :)

adamw: If needed, no problem in adding a test dep

core/src/main/scala/sttp/model/sse/ServerSentEvent.scala
magdzikk: Fixed now

core/src/main/scala/sttp/model/sse/ServerSentEvent.scala
magdzikk: Fixed - `copy` no longer re-splits, and I reversed accumulation for performance

core/src/main/scala/sttp/model/sse/ServerSentEvent.scala
magdzikk: Added property tests 👍

core/src/main/scala/sttp/model/sse/ServerSentEvent.scala
magdzikk: > It would help to add a migration target

@adamw I wonder about the name. If we keep the body as you suggested, then `ServerSentEvent.parse(List("foo: bar")).isCommentOnly` will be `true`, which might be misleading because this message doesn't really contain comments.

I guess we do want to keep the body as you suggested, so that it's a migration target for skiping keep-alives via `_ != ServerSentEvent()`. What do you think about naming it `hasNoFields` or `isSkippable`?

adamw: `hasNoFields` I think reflects the true state most directly, so let's go with that maybe?

PR #481, opened 2026-09-01
Remove line terminators in eventType/id/data

### LINKED ISSUE #480: ServerSentEvent.toString allows field injection via line terminators in eventType, id and data
`ServerSentEvent.toString` interpolates `eventType` and `id` directly, and splits `data` on `\n` only. Since the [WhatWG specification](https://html.spec.whatwg.org/multipage/server-sent-events.html) treats a CRLF pair, a lone LF **and a lone CR** as line terminators, a value containing any of these ends the current line on the wire, and the remainder is parsed by the client as a new field.

### Current behaviour

```scala
ServerSentEvent(eventType = Some("a\ndata: x")).toString
// event: a
// data: x        <- injected data field

ServerSentEvent(id = Some("a\ndata: x")).toString
// id: a
// data: x        <- injected data field

ServerSentEvent(data = Some("x\revent: foo")).toString
// data: x\revent: foo   <- a client splitting on CR sees an injected event field
```

`eventType` and `id` are the more exposed of the three: they are interpolated with no splitting or validation at all, so a plain `\n` is enough — no CR needed.

```scala
val _data = data.map(_.split("\n")).map(_.map(line => Some(s"data: $line"))).getOrElse(Array.empty[Option[String]])
val _event = eventType.map(event => s"event: $event")
val _id = id.map(id => s"id: $id")
```

This matters because `toString` is how applications render events built from their own (potentially user-supplied) values, so a caller can be made to emit fields it never intended — including `data`, `event`, `id` or `retry`.

### Suggested fix

The right remedy differs per field:

- `data` may legitimately span multiple lines, so it should split on all three terminators: `split("\r\n|\r|\n")`, each part emitted as its own `data:` line.
- `eventType` and `id` cannot span lines, so splitting is not meaningful. They should have line terminators rejected or stripped.

### Related

The same class of bug in the new `comments` field was found in review and fixed in #478 (`comments` now splits on `\r\n|\r|\n`). The three fields above are pre-existing on `master` and are untouched by that PR, hence this separate issue.

magdzikk:
Fixes #480.

`toString` could emit fields the caller never set: `eventType` and `id` were interpolated as-is, and `data` was split on `\n` only. Since the [WhatWG specification](https://html.spec.whatwg.org/multipage/server-sent-events.html) treats CRLF, a lone LF and a lone CR as line terminators, any of these in a value ended the line on the wire and the remainder was parsed by the client as a new `data`, `event`, `id` or `retry` field.

- `data` may legitimately span lines, so it now splits on all three terminators, each part emitted as its own `data:` line.
- `eventType` and `id` cannot span lines, so terminators are removed.


core/src/main/scala/sttp/model/sse/ServerSentEvent.scala
```diff
@@ -9,15 +9,22 @@ case class ServerSentEvent(
     retry: Option[Int] = None
 ) {
   override def toString: String = {
-    val _data = data.map(_.split("\n")).map(_.map(line => Some(s"data: $line"))).getOrElse(Array.empty[Option[String]])
-    val _event = eventType.map(event => s"event: $event")
-    val _id = id.map(id => s"id: $id")
+    val _data = data
+      .map(_.split(ServerSentEvent.LineTerminators))
```
adamw: `split` without a limit drops trailing empty strings, so trailing line terminators are silently lost:

- `ServerSentEvent(data = Some("\n")).toString` is `""` — the data disappears completely
- `Some("a\n")` emits just `data: a`

On master only a trailing `\n` was lost; this PR extends that to `\r` and `\r\n` (`Some("a\r")` used to round-trip intact).

`split(ServerSentEvent.LineTerminators, -1)` fixes this: `"a\n"` then emits `data: a\ndata: `, which `parse` reads back as `Some("a\n")`.

PR #490, opened 2026-09-22
Migrate to sbt 2.0.9

pierscin:
- Project's sbt is now 2.0.9
- Deleted sbt-projectmatrix
- Changed SML plugins to 3.0.0
- ~Changed CI and steward to run on JDK 25~ JDK 11 -> 17 because of change in our workflows scripts
- sbt 2 syntax adjustments
- put -release 8 explicitly in scalac options
- sbt-ide-settings: version update and org.jetbrains -> org.jetbrains.scala


.github/workflows/ci.yml
```diff
@@ -18,13 +18,15 @@ jobs:
       github.event.pull_request.head.repo.full_name != github.repository ||
       github.event.pull_request.user.login == 'softwaremill-ci'
     with:
+      java-version: "25"
```
adamw: hm why do we have to bump the java version? this should build using java 11

